PR #68, opened 2026-05-11
Skip auto-assign for bot/agent senders

### LINKED ISSUE #67: [BUG] Auto-assign-author workflow fails for bot/agent senders
**Describe the bug**
The `Auto-assign author` workflow (`.github/workflows/auto-assign-author.yml`) fails when an issue or PR is opened by a non-human sender (GitHub App, Copilot/Claude agent, etc.). The job tries to add the sender as an assignee via `github.rest.issues.addAssignees`, but GitHub's REST API rejects that call when authenticated with a GitHub App installation token, causing the workflow run to error out.

The error surfaced in the workflow log is:

```
Unhandled error: HttpError: Assigning agents is not supported with GitHub App installation tokens. Use a user token (personal access token or OAuth token) instead.
```

**To reproduce**

1. Have an automated identity (GitHub App installation or a Copilot/Claude agent) open an issue or a pull request in this repo.
2. The `Auto-assign author` workflow triggers on `issues.opened` / `pull_request.opened`.
3. The `assign-author` job fails with the `HttpError` above.
4. The issue/PR is still created, but the workflow run is marked failed and no assignee is set.

**Expected behavior**
When the sender cannot be assigned (because it is a bot or agent that the GitHub REST API refuses to assign with the installation token available to the workflow), the workflow should skip the assignment cleanly and finish successfully. Human-authored issues/PRs should continue to be auto-assigned as before.

**Screenshots**
N/A — the failure is a workflow log error rather than a UI issue.

**System information**
N/A — the bug is in a GitHub Actions workflow, not in the local Python package, so `utils/system_info.sh` output is not relevant.

**Additional context**
The fix is to detect non-`User` sender types in `actions/github-script` and return early before calling `addAssignees`, with a `try`/`catch` fallback so a future identity type that slips past the check logs a warning rather than failing the run. Assignment is a non-critical convenience — the issue/PR is created either way — so degrading gracefully is the right behavior.

**Discussion links**
Discussed directly with @zachtheyek (repo maintainer) while triaging a failed run of the workflow; filing this issue so the fix has a tracked reference and the PR can link back to it.

zachtheyek:
## Summary
- Skip the `addAssignees` call in `.github/workflows/auto-assign-author.yml` when `context.payload.sender.type !== "User"` so the workflow no longer fails when a GitHub App or Copilot/Claude agent opens an issue/PR.
- Wrap the assignment call in `try`/`catch` as a defense-in-depth fallback — if a future identity type slips past the type check, log a warning instead of failing the run.
- Collapse the duplicated `issues` / `pull_request` branches into a single `addAssignees` call now that the issue number is the only differing field.

Closes #67.

## Test plan
- [ ] Open a human-authored issue → workflow runs, issue is auto-assigned to the author (existing behavior preserved).
- [ ] Open a human-authored PR → workflow runs, PR is auto-assigned to the author (existing behavior preserved).
- [ ] Open an issue/PR via a GitHub App or agent identity → workflow finishes successfully, no assignee is set, log shows the "Skipping assignment" info line.

### DIFF
--- a/.github/workflows/auto-assign-author.yml
+++ b/.github/workflows/auto-assign-author.yml
@@ -21,22 +21,31 @@ jobs:
         uses: actions/github-script@v7
         with:
           script: |
-            const actor = context.payload.sender.login;
+            const sender = context.payload.sender;
 
-            if (context.eventName === "issues") {
-              await github.rest.issues.addAssignees({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                issue_number: context.payload.issue.number,
-                assignees: [actor],
-              });
+            // GitHub's REST API does not support assigning bot/agent accounts
+            // (e.g. GitHub Apps, Copilot/Claude agents) via App installation
+            // tokens, so skip the assignment when the sender is not a human user.
+            if (sender.type !== "User") {
+              core.info(`Skipping assignment: sender ${sender.login} has type "${sender.type}"`);
+              return;
             }
 
-            if (context.eventName === "pull_request") {
+            const issueNumber = context.eventName === "issues"
+              ? context.payload.issue.number
+              : context.payload.pull_request.number;
+
+            try {
               await github.rest.issues.addAssignees({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
-                issue_number: context.payload.pull_request.number,
-                assignees: [actor],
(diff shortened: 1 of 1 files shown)


zachtheyek (commented):
LGTM

PR #82, opened 2026-07-08
Blackwell GPU support + preflight CLI validation, docs, CI, and utils overhaul

zachtheyek:
## Summary

This branch began as Blackwell GPU support (#75) but grew to encompass several cohesive
workstreams. Each is now tracked by its own issue; this PR delivers all of them off the
single `feature/blackwell_migration` tree. **Supersedes #76**, which was scoped only to
Blackwell.

### Blackwell GPU support — Closes #75
- Dual-runtime deployment off one source tree: Ampere via the existing conda env, Blackwell via the NGC TF 25.02 container, both at TF 2.17.
- Runtime-agnostic container wrapper (auto-detects `apptainer` vs `singularity`); `aetherscan.def` + `requirements-container.txt`.
- Lifts hardcoded GPU-strategy values into `GPUConfig` + CLI flags (`--gpu-memory-limit-mb`, `--nccl-num-packs`, `--async-allocator`/`--no-async-allocator`); NCCL fallback now wraps a real warmup all-reduce.
- Keras 3 / TF 2.17 modernization in `Sampling` and `setup_gpu_strategy` (`tf.random.normal`, `register_keras_serializable`, `set_logical_device_configuration`).

### Preflight CLI validation + centralized config/GPU selection — Closes #77
- `validate_args` + `collect_validation_errors`: structured, mode-gated bounds / cross-parameter / cross-replica divisibility checks run before `apply_args_to_config`.
- Reusable `_add_train_flags_to` / `_add_inference_flags_to`; `--num-replicas` wired through `GPUConfig`; `_resolve_num_replicas` centralizes replica-count resolution + hardware validation.
- `--config-path` loads a saved config onto the singleton before validation; fix-suggestion surface shared with `find_optimal_configs.py`; documented in `docs/CONFIG_AND_CLI.md`.

### Docstring normalization to plain prose — Closes #78
- Converts module/function docstrings across the codebase to the plain-prose style prescribed by `CONTRIBUTING.md`. No behavior change.

### Claude CI/automation workflows — Closes #79
- update-docs (README CLI reference + lean-CLAUDE.md/skill sync), code-review, issue-triage, contribution-check, release-notes, weekly dependency-update and flaky-test trackers.
- Standardized prompts, documented tool allowlists, `timeout-minutes` caps. (These are the workflows referenced by #59.)

### Contributor docs restructuring — Closes #80
[...]

PR #91, opened 2026-07-11
Add manual workflow_dispatch re-scan to claude-update-docs

### LINKED ISSUE #90: Add a manual re-scan trigger (workflow_dispatch) to claude-update-docs
## Summary

`claude-update-docs` only triggers on a real `pull_request: closed` (merged) event, so there's no way to run it against an **already-merged** PR on demand.

This matters because GitHub runs `pull_request` workflows using the workflow file **from the triggering commit**, not the latest default branch. So a plain "re-run" of an old PR's merge event replays that commit's *older* workflow file — it can't be used to validate or apply newer update-docs logic against a historical PR.

## Proposal

Add a `workflow_dispatch` trigger with a `pr_number` input. `workflow_dispatch` runs use the default-branch workflow file, so this gives a clean "re-scan an old PR with the current logic" path. Route both triggers through a single job-level `PR_NUMBER` so the merge-path behavior is unchanged.

## Immediate use

Validate the #89 fixes (see #88) by re-scanning PR #82 — which touches both `cli.py` (contract (a)) and the canonical docs (contract (b)) — and confirming the filed issue is actionable and doesn't double-trigger contribution-check.

## Discussion links

Scoped directly with the maintainer (@zachtheyek) in a working session; no separate GitHub Discussion.

zachtheyek:
Closes #90

## What & why
Adds a manual `workflow_dispatch(pr_number)` trigger to `claude-update-docs` so the **current** workflow logic can be run against an already-merged PR. A plain re-run of a past merge replays that commit's older workflow file; `workflow_dispatch` runs use the default-branch file, so this is the reliable re-scan path.

- `workflow_dispatch` trigger with a required `pr_number` input.
- Job runs on `workflow_dispatch` as well as a merged `pull_request`.
- Single job-level `PR_NUMBER` env (`inputs.pr_number` on dispatch, else the merged PR) drives every step and the prompt.
- On dispatch the event has no PR title/body/author/merge sha, so the prompt fetches them via `gh pr view` when blank.

The merge-triggered path is unchanged.

## Verification
- `yaml.safe_load` parses; `workflow_dispatch.inputs.pr_number` present; job `if` allows both triggers; every PR-number site now uses `env.PR_NUMBER`.
- pre-commit (incl. check-yaml) passes; commit GPG-signed.

## Note
Merging this will itself fire `claude-update-docs` once on this PR's own (workflow-only) diff — expected. The real test is the follow-up manual dispatch against #82.

### DIFF
--- a/.github/workflows/claude-update-docs.yml
+++ b/.github/workflows/claude-update-docs.yml
@@ -6,11 +6,22 @@ on:
   pull_request:
     types: [closed]
     branches: [master]
+  # Manual re-scan of an already-merged PR. A workflow_dispatch run uses the workflow file from
+  # the default branch, so this is the way to run the CURRENT logic against an old PR — a plain
+  # re-run of a past merge would replay that commit's older workflow file instead.
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "Number of a merged PR to (re)scan for documentation updates"
+        required: true
+        type: string
 
 jobs:
   claude-update-docs:
-    # Only run when PR is actually merged, not just closed
-    if: github.event.pull_request.merged == true
+    # Run on a real merge to master, or a manual workflow_dispatch re-scan of a merged PR.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.pull_request.merged == true
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


zachtheyek:
comments addressed. LGTM

PR #100, opened 2026-07-11
Fix claude-update-docs follow-up loop (bot allowlist, CI noise, CLI embedding)

### LINKED ISSUE #99: Fix claude-update-docs follow-up loop: bot allowlist, contribution-check noise, CLI-block embedding
## Summary

Testing the `claude-update-docs` re-scan against #82 (issue #97) validated the pipeline end-to-end but surfaced three defects in the follow-up half of the loop.

## Defects & fixes

1. **Follow-up `@claude` PR never fires on the auto-filed issue.** The doc issue is authored by bot **`claude`** (`app/claude`, type Bot), but `claude.yml`'s `allowed_bots` was `"claude[bot]"` — which doesn't match — so the assistant silently skipped it. → Allow `"claude,claude[bot]"`.

2. **`claude-contribution-check` leaves a red "failure" on every bot-authored issue.** With `allowed_bots` unset, the action correctly refuses bots, but the job still spins up and `claude-code-action` aborts with a "non-human actor" error → a failed run (no comment, but noisy). → Re-add a job-level `if:` guard that skips the whole job for bot/automation actors.

3. **Contract (a) hands off a file that doesn't exist downstream.** The update-docs Claude step *referenced* `cli_help_regenerated.txt` instead of pasting the blocks into the issue. That file is ephemeral to the update-docs runner, so the follow-up PR (fresh checkout) can't read it. → Add a shell step that appends the regenerated blocks to the filed issue straight from the file (guaranteed verbatim), and stop the prompt from pasting/referencing the path.

## Discussion links

Found while testing #88/#90 against #82 with the maintainer; no separate GitHub Discussion.

zachtheyek:
Closes #99

Three fixes surfaced by re-scanning #82 (issue #97), one commit each:

- **claude.yml** — `allowed_bots` `"claude[bot]"` → `"claude,claude[bot]"`. The auto-filed doc issue is authored by bot `claude` (`app/claude`), so the follow-up assistant was silently skipped.
- **claude-contribution-check.yml** — re-add the job-level bot-skip `if:` guard, so it no longer spins up a runner and shows a red "failure" on every bot-authored issue.
- **claude-update-docs.yml** — add a deterministic step that appends the regenerated CLI blocks to the filed issue straight from `cli_help_regenerated.txt` (the follow-up can't read that ephemeral file), and update the prompt so Claude neither pastes nor references it.

## Verification
- All three YAMLs parse; contribution-check has the guard; update-docs has the new "Embed regenerated CLI Reference" step; claude.yml `allowed_bots` = `claude,claude[bot]`.
- Simulated the append step's body-building locally — well-formed markdown, balanced code fences.
- pre-commit (incl. check-yaml) passes; all commits GPG-signed.

After merge, re-testing #82 needs #97's `pr=82` marker stripped first (else the dedup guard skips) — I'll handle that.

### DIFF
--- a/.github/workflows/claude-contribution-check.yml
+++ b/.github/workflows/claude-contribution-check.yml
@@ -16,6 +16,16 @@ permissions:
 
 jobs:
   contribution-check:
+    # Never run on bot- or automation-authored issues/PRs. Our own workflows (e.g.
+    # claude-update-docs) file issues as the `claude` bot; without this guard the job would spin
+    # up a runner and claude-code-action would abort with a "non-human actor" error — a red
+    # failure on every bot issue. Skip the whole job instead. (allowed_bots is already unset, so
+    # no comment is ever posted; this additionally avoids the noisy failed run.) See issue #83.
+    if: >-
+      github.event.issue.user.type != 'Bot' &&
+      github.event.pull_request.user.type != 'Bot' &&
+      !endsWith(github.actor, '[bot]') &&
+      github.actor != 'claude'
     runs-on: ubuntu-latest
     # Contribution check shouldn't take more than 5 mins (refine if data shows otherwise)
     timeout-minutes: 5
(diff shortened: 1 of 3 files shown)


zachtheyek:
LGTM

PR #200, opened 2026-07-21
ci: SHA-pin all workflow actions + claude-code-action v1.0.102 + pre-commit hook bumps

### LINKED ISSUE #59: Node.js 20 deprecation warnings in Claude workflows (post-June 2 follow-up)
## Context

After updating `claude-code-action` to v1.0.85 (#56, #57), the Claude workflows went from **errors to warnings**, which is acceptable for now. However, the Node.js 20 deprecation warnings persist because `claude-code-action` internally depends on `oven-sh/setup-bun`, which still declares `node20` as its runtime.

## Warning

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: oven-sh/setup-bun@3d267786b128fe76c2f16a390aa2448b815359f3.

## Timeline

- **June 2, 2026**: GitHub forces all actions to Node.js 24 by default — warnings may become errors again
- **September 16, 2026**: Node.js 20 removed from runners entirely

## Affected workflows

- `claude.yml`
- `claude-code-review.yml`
- `claude-contribution-check.yml`
- `claude-issue-triage.yml`
- `claude-update-docs.yml`

## Action items (after June 2)

1. Check if a newer `claude-code-action` release ships with a Node.js 24-compatible `setup-bun` — if so, update the pinned commit
2. If not, add `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` as a job-level env var to all 5 workflows to override the runtime
3. Alternatively, wait for upstream to fix and monitor for breakage

## References

- Previous fix: #56 / #57
- GitHub deprecation notice: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### LINKED ISSUE #140: Test/CI hardening: SHA-pin GitHub Actions + xdist-safe global monkeypatch
## Context

Two non-blocking observations from the second-pass code review of the test-suite PR (#116). Both are low-priority hardening items deliberately deferred out of that PR and tracked here so they aren't lost in review threads.

## 1. GitHub Actions pinned by tag, not commit SHA

**Problem.** Workflows reference actions by moving tags (e.g. `actions/checkout@v5`, `actions/setup-python@v5`, and others across `.github/workflows/`). A moving tag is a supply-chain vector: if a tag is repointed to a malicious commit, CI would execute it. `permissions: contents: read` on the tests workflow limits blast radius, but not every workflow is equally scoped.

**Correct solution.** Pin third-party actions to a full commit SHA with the human-readable version in a trailing comment, e.g. `uses: actions/checkout@<40-char-sha> # v5.0.0`. Apply repo-wide (all workflows in `.github/workflows/`), and enable Dependabot's `github-actions` ecosystem to keep the pinned SHAs updated. First-party `actions/*` are lower-risk, but pinning them too is the consistent, auditable choice.

**Scope.** Repo-wide, one deliberate pass — not piecemeal per-PR.

## 2. Global `np.random.choice` monkeypatch is xdist-fragile

**Problem.** `test_near_zero_slope_is_clamped` (in the `data_generation` unit tests) monkeypatches `np.random.choice` process-globally. It is safe today — tests run serially and `monkeypatch` restores the original after the test — but under parallel execution with `pytest-xdist`, a process-global patch shared across tests running concurrently in the same worker could bleed or race.

**Correct solution.** If/when `pytest-xdist` is adopted: scope the patch tightly — patch the specific reference the code under test resolves (e.g. the `np.random.choice` symbol as imported into `data_generation`) instead of the global `numpy.random.choice`, or inject a seeded `numpy.random.Generator` and have the code accept it. Until xdist is adopted there is no action needed; this records the constraint so a future parallelization effort doesn't get silently flaky results.

**Scope.** Only relevant if parallel test execution is introduced.

## Priority

Both low; neither blocks anything today. Filed for tracking.

zachtheyek:
Closes #59
Closes #140

## What changed

One workflows-hardening pass, three focused commits:

**1. SHA-pin every remaining tag-pinned action (#140).** All 27 tag-pinned `uses:` lines across the 14 files in `.github/workflows/` now carry a full commit SHA plus a trailing `# vX.Y.Z` comment:

| Action | Old ref | New pin | Count |
|---|---|---|---|
| actions/checkout | `@v5` | `fbc6f399...` # v5.1.0 | 14 |
| actions/setup-python | `@v5` | `a26af69b...` # v5.6.0 | 5 |
| actions/cache | `@v3` | `0057852b...` # v4.3.0 | 2 |
| actions/upload-artifact | `@v4` | `ea165f8d...` # v4.6.2 | 1 |
| actions/download-artifact | `@v4` | `d3f86a10...` # v4.3.0 | 2 |
| actions/github-script | `@v7` | `f28e40c7...` # v7.1.0 | 3 |

Every SHA was resolved via the GitHub API and verified to be the exact commit its named release tag points to. For everything except `actions/cache`, the pinned commit is what the moving major tag resolves to **today**, so this freezes current CI behavior rather than changing any version. The local reusable-workflow reference (`release.yml` -> `./.github/workflows/tests.yml`) is not pinnable and is unchanged; `pre-commit/action` and `pypa/gh-action-pypi-publish` were already SHA-pinned.

**2. claude-code-action v1.0.85 -> v1.0.102 (#59).** All 9 workflows using `anthropics/claude-code-action` move from `58dbe8ed...` # v1.0.85 to `5d5c10a4...` # v1.0.102. v1.0.85 vendors `oven-sh/setup-bun` v2.1.2 (`using: node20`), which fires the Node.js 20 deprecation annotation on every Claude workflow run. v1.0.102 (2026-04-20) is the first release shipping setup-bun v2.2.0 — verified by reading `action.yml` at the pinned commit (pins setup-bun `0c5077e5`) and confirming that setup-bun commit declares `using: "node24"`. The issue's fallback (`FORCE_JAVASCRIPT_ACTIONS_TO_NODE24`) is deliberately **not** added: runners already force Node 24 and the workflows succeed, so the env var is obsolete — this bump removes the warning noise at the source. Also resolves the identical recommendation in the weekly dependency-check issue #174.

**3. Pre-commit hook bumps (deferred from #188).** `.pre-commit-config.yaml`: `astral-sh/ruff-pre-commit` v0.14.4 -> v0.14.14 and `gitleaks` v8.29.0 -> v8.29.1 (both patch-level). `pre-commit run --all-files` passes with the new versions — no new lint findings, no autofixes, so no source changes ride along. (ruff-pre-commit now reports the `ruff` hook id as a legacy alias of `ruff-check`; it still works, left as-is to keep the diff minimal.)

## Why no dependabot.yml

The pin-freshness role #140 assigned to Dependabot is already covered: the weekly `claude-dependency-check.yml` explicitly audits `.github/workflows/*.yml` action pins (it is what produced the v1.0.150 recommendation in #174). Adding Dependabot's `github-actions` ecosystem would duplicate that with a second bot — matching the owner's 2026-07-21 comment on #140.

## xdist monkeypatch item (#140, item 2)

Stays dormant by design, per the issue's own scope statement: pytest-xdist is not adopted anywhere (dev deps and CI run plain serial pytest), and the patch in `test_near_zero_slope_is_clamped` already targets the exact module-attribute reference the code under test resolves. No action until thread-level test parallelism is ever introduced; nothing in this PR touches it.

## Maintainer decisions applied

- **actions/cache v3 -> v4 major bump** (pre-commit.yml, 2 steps): a rider on the pinning pass rather than a pure freeze. v3 is an aging major; v4.3.0 (2025-09-24) is ~10 months seasoned, well past SECURITY.md's bar. Both cache steps use only `path`/`key` inputs, which are unchanged in v4. Veto by re-pinning to `6f8efc29...` (v3's current target) if preferred.
- **claude-code-action target = v1.0.102, not v1.0.150 (#174's rec) or latest (v1.0.179)**: no release both clears the node20 deprecation (>= v1.0.102, first shipped 2026-04-20) and meets the normal 6-month lag (<= 2026-01-22) — the constraint set is empty. Per SECURITY.md's advisory-override case, the deprecation on the current pin overrides the lag; v1.0.102 is the oldest release clearing it (~3 months in the wild, the most seasoning available) and the smallest behavioral jump from v1.0.85 (17 patch releases vs 65+). Residual risk either way: any bump across this range changes the action's own behavior — worth watching the first few claude-code-review/claude.yml runs.
- **Pinned first-party actions at the moving tag's current target** (e.g. checkout v5.1.0, released 2026-07-20): CI already runs these commits via the moving tags, so freezing them is the no-behavior-change choice even where the release is recent.

## Verification

- `grep -E "uses: " .github/workflows/*.yml` shows every external action pinned to a 40-char SHA with a version comment; the only non-SHA `uses:` is the local reusable `./.github/workflows/tests.yml`.
- Every SHA cross-checked against its release tag via `gh api` (annotated tags dereferenced to commits — the claude-code-action tag object dereferences to commit `5d5c10a4...`).
- `pre-commit run --all-files` green locally with the bumped hooks (includes `check-yaml` over all edited workflow files, plus ruff lint+format repo-wide).
- No Python code changed, so no pytest run applies; CI on this PR is the real proof that every pin resolves, and the absence of the Node.js deprecation annotation on this PR's own Claude workflow runs verifies #59.

### DIFF
--- a/.github/workflows/auto-assign-author.yml
+++ b/.github/workflows/auto-assign-author.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: Assign author to issue/PR
-        uses: actions/github-script@v7
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
         with:
           script: |
             const sender = context.payload.sender;
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Run Claude Code Review
         id: claude-review
-        uses: anthropics/claude-code-action@58dbe8ed6879f0d3b02ac295b20d5fdfe7733e0c # v1.0.85
+        uses: anthropics/claude-code-action@5d5c10a4f389689f992ea10bb14dcb6fcc83146d # v1.0.102
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           allowed_bots: "claude[bot]"
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
           fetch-depth: 1
 
--- a/.github/workflows/claude-contribution-check.yml
+++ b/.github/workflows/claude-contribution-check.yml
@@ -37,7 +37,7 @@ jobs:
           fetch-depth: 1
 
       - name: Run Claude Contribution Check
-        uses: anthropics/claude-code-action@58dbe8ed6879f0d3b02ac295b20d5fdfe7733e0c # v1.0.85
+        uses: anthropics/claude-code-action@5d5c10a4f389689f992ea10bb14dcb6fcc83146d # v1.0.102
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: |
--- a/.github/workflows/claude-contribution-check.yml
+++ b/.github/workflows/claude-contribution-check.yml
@@ -32,7 +32,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
           fetch-depth: 1
 
--- a/.github/workflows/claude-dependency-check.yml
+++ b/.github/workflows/claude-dependency-check.yml
@@ -49,7 +49,7 @@ jobs:
 
       - name: Checkout repository
         if: steps.guard.outputs.already_filed != 'true'
-        uses: actions/checkout@v5
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
         with:
           # Manifests are read from the working tree; no git history needed.
           fetch-depth: 1
--- a/.github/workflows/claude-dependency-check.yml
+++ b/.github/workflows/claude-dependency-check.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: Run Claude Dependency Check
         if: steps.guard.outputs.already_filed != 'true'
-        uses: anthropics/claude-code-action@58dbe8ed6879f0d3b02ac295b20d5fdfe7733e0c # v1.0.85
+        uses: anthropics/claude-code-action@5d5c10a4f389689f992ea10bb14dcb6fcc83146d # v1.0.102
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: |
--- a/.github/workflows/claude-flaky-test-tracker.yml
+++ b/.github/workflows/claude-flaky-test-tracker.yml
@@ -56,7 +56,7 @@ jobs:
 
       - name: Run Claude Flaky Test Tracker
         if: steps.guard.outputs.already_filed != 'true'
-        uses: anthropics/claude-code-action@58dbe8ed6879f0d3b02ac295b20d5fdfe7733e0c # v1.0.85
+        uses: anthropics/claude-code-action@5d5c10a4f389689f992ea10bb14dcb6fcc83146d # v1.0.102
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: |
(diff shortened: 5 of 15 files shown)


zachtheyek:
## Why `claude-code-review` and `contribution-check` are red on this PR (by design)

Both failing checks are the claude-code-action's anti-tamper guard, not a defect in this PR. Their logs show the action failing during app-token exchange, before any review work starts (`gh run view --log-failed`, runs 29867554177 / 29867553906):

> App token exchange failed: 401 Unauthorized - Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch. If you're seeing this on a PR when you first add a code review workflow file to your repository, this is normal and you should ignore this error.

`pull_request`-triggered jobs execute the workflow file **from the PR branch**, and this PR's entire purpose is to modify those files — every one of the 14 workflows, including `claude-code-review.yml` and `claude-contribution-check.yml` themselves (new SHA pins + the claude-code-action v1.0.85 -> v1.0.102 bump). So the PR-branch copy can never be "identical content to the version on the repository's default branch", and these two checks **cannot pass pre-merge** on any PR that touches these workflows. The action's own log message says to ignore the error in this situation. After merge, subsequent PRs will run the (then-default-branch) updated workflows normally.

Everything that can pass is green on the head commit: **tests (3.10, 3.12), pre-commit, and CodeQL (python + actions)** all succeed. Note for the maintainer: if either red check is marked required in branch protection, merging will need an admin override/bypass for this one PR.

Since the `pull_request` path can't produce the automated first-pass review here, requesting it via the comment path instead — `issue_comment`-triggered runs use `claude.yml` from the default branch, so the guard doesn't apply.

@claude please review this PR: validate the SHA pins against their stated release tags, the claude-code-action v1.0.102 bump rationale, and the pre-commit hook bumps.

PR #204, opened 2026-07-21
feat: persist RF eval metrics to the DB + live RF dashboard tab

### LINKED ISSUE #171: Persist RF training/eval metrics to the DB for live RF monitoring on the dashboard
### Motivation
The live dashboard (#170) reconstructs beta-VAE curves, injection stats, latent snapshots, resources, stage timeline, and inference candidates directly from the DB. But **nothing about the Random Forest is written to any DB table** — every RF diagnostic (`train.py` `plot_rf_*`) is sourced from the on-disk `{model_path}/rf_eval_artifacts_{tag}.joblib` (+ the SHAP cache + the live model), never `training_stats`/`injection_stats`/etc. (all `write_training_stat` calls use `model_name="beta_vae"`).

Consequence: RF plots can only appear in the dashboard's **PNG gallery** as saved artifacts, not as live interactive charts — and RF progress can't be monitored live at all the way beta-VAE loss can. This asks to persist the RF metrics to the DB so RF training/eval becomes first-class in the live dashboard (and the benchmark/report tooling).

### Proposal
Write RF eval metrics to the DB when `rf_eval_artifacts_{tag}.joblib` is produced. Either extend `training_stats` with `model_name="rf"` rows, or add a small dedicated table. Candidate scalars/series to persist (all already computed for the existing plots — see the inventory in the #134 review):
- Binary + sub-type validation accuracy, ROC-AUC, PR-AP, Brier score (calibration).
- Ensemble accuracy vs tree count (the OOB/accuracy-elbow curve).
- Confusion-matrix cell counts (binary 2×2 + sub-type 4×2).
- Confidence-distribution summary (histogram bins or quantiles of val P(true)).
- Optionally: top-k feature importances / mean|SHAP| per feature.

Then add an "RF" tab to `utils/dashboard.py` that renders these live (mirroring the beta-VAE tab), and keep the richer figures (SHAP beeswarm, decision boundary) in the PNG gallery since they need arrays/models.

### Non-goals
- Persisting full SHAP value matrices or per-sample probas (large; the joblib + PNG gallery already cover those).

Surfaced by the plot inventory during the #170 dashboard work (RF was the one plot family with no DB backing).

zachtheyek:
Closes #171

## What changed

The Random Forest was the one model family with no DB backing — every RF diagnostic was sourced from the on-disk `rf_eval_artifacts_{tag}.joblib` and appeared on the dashboard only via the PNG gallery. This PR persists RF eval metrics to `training_stats` and gives the dashboard a first-class RF tab.

- **`src/aetherscan/rf_metrics.py` (new)** — `compute_rf_eval_metrics()`, a pure helper (arrays in → `{stat_name: float}` out) computing val accuracy / ROC-AUC / average precision / Brier score, per-sub-type accuracies, binary 2×2 + sub-type × prediction confusion cell counts, and confidence quantiles of val P(true). Deliberately a top-level TF-free module (the `models` package `__init__` pulls TensorFlow via `vae.py`), so it is unit-testable without the GPU stack.
- **`train.py` `train_random_forest()` tail** — calls the helper on the already-in-memory val arrays and writes each scalar via `db.write_training_stat(model_name="rf", ...)`, plus a `classification_threshold` row. Writes happen only after both RF joblibs land, so any later retry resumes via `try_load_rf_for_resume()` and skips retraining — the rows are never duplicated by the resume path. The block is best-effort: a metrics failure logs a warning and never fails the training run.
- **`train.py` `plot_rf_ensemble_accuracy_curve()`** — writes the per-tree `ensemble_val_accuracy` series (`epoch_number` = tree count) from inside the plot function where the O(n_trees) per-tree `predict_proba` loop already runs, rather than duplicating that cost at train time. A comment notes the resulting dependency: the series only lands when the `rf_plots` stage runs and this plot succeeds, and it still thresholds at the pre-existing hard-coded 0.5 (unlike the scalar `val_accuracy`, which uses the deployment threshold — the stat names + an in-tab caption disambiguate).
- **`dashboard.py`** — `load_rf_stats()` (alive `model_name='rf'` rows with last-write-wins dedup on `(stat_name, epoch_number)`, which absorbs rf_plots-retry re-writes and reused-tag stale scalars), `rf_confusion_matrices()` pure shaping helper, and an **RF** tab: metric tiles (accuracy/AUC/AP/Brier), binary + sub-type confusion heatmaps, per-sub-type accuracy bars, the ensemble-accuracy-vs-tree-count curve, and confidence quantiles. SHAP / decision-boundary / calibration figures stay in the PNG gallery as planned. The dashboard stays standalone (stdlib + numpy/pandas/plotly/streamlit only).
- **Tests** — `tests/unit/test_rf_metrics.py` (hand-computed AUC/AP/Brier/confusion/quantile values, perfect-prediction case, float-purity + key-count) and new `test_dashboard.py` cases (seeded-sqlite `load_rf_stats` model/superseded filtering + dedup semantics, `rf_confusion_matrices` shaping, empty-tag behavior).

Note: the issue body says `utils/dashboard.py`; the actual dashboard is `src/aetherscan/dashboard.py` (there is no utils dashboard), which is what this PR edits.

## Verification

- `tests/unit/test_rf_metrics.py` + `tests/unit/test_dashboard.py`: **27 passed** locally in the TF venv (`-m "not gpu and not cluster"`).
- Adjacent suites `test_dashboard_launcher.py`, `test_train_utils.py`, `test_train_datasets.py`: **65 passed** locally.
- `import aetherscan.train` verified in the TF venv (new import wiring).
- `ruff check` + `ruff format` clean on all changed files (pre-commit hooks passed).
- Full suite deferred to CI. The train-side write paths (train_random_forest tail, plot fn) are not exercised by unit tests — they need a GPU/cluster run, matching the existing coverage posture for `train.py`'s beta-VAE DB writes.

## Maintainer decisions applied

- **Reuse `training_stats` (no schema v5)** — rows go in with `model_name='rf'`, which `write_training_stat`'s docstring already anticipated; the ensemble curve's x-value is carried in `epoch_number` (= tree count). A dedicated typed `rf_metrics` table would need a v5 bump + writer/query plumbing; veto if you prefer that.
- **Split computation: scalars at train time, ensemble curve at plot time** — the curve series reuses the expensive per-tree loop inside `plot_rf_ensemble_accuracy_curve`, so those rows depend on the `rf_plots` stage succeeding (scalars do not).
- **Duplicate handling is read-side** — `load_rf_stats` dedupes last-write-wins instead of adding a model/stat-scoped `mark_superseded` variant (the existing supersede mark is `round_ge`-based and RF rows carry no round number; a write-side mark would have superseded beta-VAE rows too).
- **Threshold disambiguation via naming + caption, not code change** — the scalar `val_accuracy` uses the deployment `inference.classification_threshold` (persisted as its own row); `ensemble_val_accuracy` keeps the plot's pre-existing hard-coded 0.5 (flagged by the existing NOTE there, unchanged here).
- **Feature importances / mean-|SHAP| left out** — the issue marked them optional; the SHAP figures remain in the PNG gallery.
- **Metric persistence is best-effort** — wrapped in try/except with a warning so an sklearn edge case (e.g. degenerate single-class val split) can't fail a multi-hour training run.

### DIFF
--- a/src/aetherscan/dashboard.py
+++ b/src/aetherscan/dashboard.py
@@ -3,10 +3,11 @@
 Live dashboard for an Aetherscan pipeline run, read straight from its SQLite DB.
 
 Renders, in one auto-refreshing page, every run metric that is reconstructable from the DB —
-resource utilization (CPU/RAM/GPU), beta-VAE loss + stability curves, the signal-injection
-stats suite, a live latent-space scatter, the stage timeline (PR #134), and inference candidate
-stats — plus a gallery of every saved plot PNG (RF diagnostics, latent traversals, inference
-figures) as it appears on disk.
+resource utilization (CPU/RAM/GPU), beta-VAE loss + stability curves, RF eval metrics
+(accuracy/AUC/AP/Brier tiles, confusion heatmaps, ensemble curve, confidence quantiles), the
+signal-injection stats suite, a live latent-space scatter, the stage timeline (PR #134), and
+inference candidate stats — plus a gallery of every saved plot PNG (RF diagnostics, latent
+traversals, inference figures) as it appears on disk.
 
 It is STANDALONE — no `aetherscan` imports (like utils/benchmark_report.py), only stdlib sqlite3 +
 numpy + pandas + plotly + streamlit — so streamlit can execute it directly and it runs against a
@@ -114,6 +115,58 @@ def load_training_stats(conn: sqlite3.Connection, tag: str, stats: list[str]) ->
         params=(tag, *wanted),
     )
     return df
+
+
+def load_rf_stats(conn: sqlite3.Connection, tag: str) -> pd.DataFrame:
+    """Non-superseded model_name='rf' training_stats for `tag` (timestamp, stat_name, value,
+    round_number, epoch_number). Duplicate (stat_name, epoch_number) rows — an rf_plots retry
+    re-writing the ensemble curve, or a reused tag's stale scalars — collapse last-write-wins
+    (rows are ordered by timestamp; scalar rows share a NULL epoch_number, which pandas
+    treats as equal when de-duplicating)."""
+    df = pd.read_sql_query(
+        "SELECT timestamp, stat_name, value, round_number, epoch_number "
+        f"FROM training_stats WHERE tag = ? AND model_name = 'rf' AND {_ALIVE} "
+        "ORDER BY timestamp",
+        conn,
+        params=(tag,),
+    )
+    if not df.empty:
+        df = df.drop_duplicates(subset=["stat_name", "epoch_number"], keep="last").reset_index(
+            drop=True
+        )
+    return df
+
+
+def rf_confusion_matrices(rf: pd.DataFrame) -> tuple[pd.DataFrame | None, pd.DataFrame | None]:
+    """Shape load_rf_stats rows into the RF tab's two confusion matrices: the binary 2x2
+    (actual x predicted, from confusion_{tn,fp,fn,tp}) and the sub-type x predicted-class
+    counts (from confusion_{subtype}_pred_{false,true}). Either is None when its cell rows
+    are absent."""
+    if rf.empty:
+        return None, None
+    scalars = dict(zip(rf["stat_name"], rf["value"], strict=False))
+
+    cells = [scalars.get(f"confusion_{c}") for c in ("tn", "fp", "fn", "tp")]
+    binary = None
+    if None not in cells:
+        binary = pd.DataFrame(
+            [[cells[0], cells[1]], [cells[2], cells[3]]],
+            index=["actual false", "actual true"],
+            columns=["pred false", "pred true"],
+        )
+
+    subtype = None
+    sub = rf[rf["stat_name"].str.match(r"^confusion_.+_pred_(true|false)$")].copy()
+    if not sub.empty:
+        parts = sub["stat_name"].str.extract(r"^confusion_(.+)_pred_(true|false)$")
+        sub["subtype"] = parts[0]
+        sub["pred"] = "pred " + parts[1]
+        subtype = (
+            sub.pivot_table(index="subtype", columns="pred", values="value", aggfunc="last")
+            .reindex(columns=["pred false", "pred true"])
+            .sort_index()
+        )
+    return binary, subtype
 
 
 def load_injection_stats(conn: sqlite3.Connection, tag: str) -> pd.DataFrame:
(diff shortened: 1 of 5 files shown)


zachtheyek:
## Review response (round 1)

### Addressed

**Single-class val split → per-metric degradation** (inline note on `rf_metrics.py`) — agreed, fixed in 7f3c074. Instead of catching `ValueError` per call, `compute_rf_eval_metrics()` now guards on class count (`np.unique(val_binary_labels).size > 1`). That choice matters because the two ranking metrics fail differently: `roc_auc_score` raises, but `average_precision_score` does **not** — on sklearn 1.5.2 an all-negative split just emits a `UserWarning` and returns `-0.0`, a silently-garbage value that a `try/except ValueError` would have happily persisted. With one class present, both keys are omitted (with a warning logged); everything else — accuracy, Brier, confusion cells, sub-type accuracies, quantiles — still lands. No dashboard change needed: the tile loop already reads `scalars.get(name)` and renders absent stats as an em dash. Docstring updated, plus a new hand-computed unit test (`test_single_class_split_omits_ranking_metrics`). Ran `tests/unit/test_rf_metrics.py`, `test_dashboard.py`, and `test_dashboard_launcher.py` locally in the TF venv: 39 passed. (Train-side callers unchanged; the blanket best-effort guard in `train_random_forest()` stays as the outer safety net.)

### Not changed, with reasons

**Wrap the ensemble-curve DB writes in try/except** (inline note on `train.py:6016-6025`) — the premise doesn't hold, so I'm leaving the loop bare. `write_training_stat` never touches SQLite at the call site: it is a non-blocking `self.write_queue.put(...)` onto an *unbounded* `Queue()` (`db/db.py:996`, queue built at `db/db.py:155`), preceded only by `get_system_metadata()` (hostname/user/pid; its socket probe swallows its own errors). The cited failures — DB locked, disk full — can only occur in the background writer thread, which already has log-and-continue semantics at two levels: `_flush_buffer`'s own try/except and `_writer_loop`'s catch-all (`db/db.py:727-730`). They cannot propagate into `plot_rf_ensemble_accuracy_curve`. This is also why every pre-existing `write_training_stat` call in `train.py` is bare, including the per-epoch beta-VAE loss/gradient/LR loops (~lines 1435-1514) — the new loop follows the established pattern. And the try/except around the scalar block in `train_random_forest()` isn't evidence to the contrary: it guards `compute_rf_eval_metrics()`'s real sklearn computation (the thing that genuinely raised, per the fix above), not the queue puts. Wrapping a queue put would be defensive code for a failure mode the DB layer already makes unreachable.

Observations 4-11: noted, no action taken. In particular, #5's n_trees write volume goes through the same non-blocking queue put (no train-loop stall possible), and #11's partial-binary-cells `None` → skipped-heatmap behavior is intentional and covered implicitly by the render guard.

@claude please re-review.

PR #249, opened 2026-07-22
test: coverage hardening — CI marker leak-guard, #221/#217/#203/#207 regression tests, smoke model-tag override

### LINKED ISSUE #248: Test-coverage hardening: CI marker leak-guard, #221/#207/#203 regression tests, smoke model-tag override
A wholesale audit of the test suite (unit + integration + markers/conftest + CI) found it mature and healthy (797 passed / 3 documented skips, disciplined markers, no xfail/TODO markers). No must-fix gaps. This issue tracks the high-value, cheap improvements worth landing; the rest are accepted as documented cluster-only or observability-only gaps.

## Add
- **CI marker leak-guard** (`.github/workflows/tests.yml`): change the selection to `pytest -m "not gpu and not cluster and not integration"` so an `integration`-only test (which skips the singleton-isolation fixture) can never leak into CI regardless of co-marking.
- **`logger.stop()` regression test** (`tests/unit/test_logger.py`): assert `stop()` calls both `log_queue.close()` and `log_queue.cancel_join_thread()` — converts the #221 interpreter-hang (currently only caught as a >90s suite timeout) into a fast, localized failure.
- **`_grouped_legend_entries` unit test** (`tests/unit/test_monitor.py`): parametrized over GPU counts vs `max_rows`, asserting `ncol` and the column-major group padding (the #217 pure helper is currently untested).
- **benchmark-report smoke assertion** (`tests/integration/test_train_smoke.py`, mirror in `test_inference_smoke.py`): after the existing config-snapshot assert, `assert os.path.exists(.../plots/benchmark_report_{tag}.png)` — pins the #203 hook's real db.flush→render path end-to-end at zero extra runtime.
- **`smoke_model_tag` fixture** (`tests/integration/conftest.py`): `os.environ.get("AETHERSCAN_SMOKE_MODEL_TAG", "test_v17")`, consumed by `test_inference_smoke.py` + `test_model_behavior.py` in place of the hardcoded `test_v17`, so the consumer smokes can point at whatever tag `test_train_smoke` just produced (they otherwise always hit their skip-guards on a fresh cluster).
- **seeding-determinism cluster smoke** (`tests/integration/test_seeding_determinism.py`, marked `integration,gpu,cluster`): mirror `test_model_behavior.py`'s in-process pattern — under `strategy.scope()`, set the seed + `enable_op_determinism()`, build the beta-VAE twice, one train step on identical synthetic snippets, assert byte-identical `get_weights()`. This is the only coverage of #207's `--tf-deterministic-ops` promise.

## Document as accepted gaps (`docs/TESTING.md`)
- `slack_handler.py` has no unit tests (observability-only; note the cooldown/backoff/color helpers are network-free and cheaply testable if hardened later).
- End-to-end training-loop resume and clean SIGTERM/SIGINT shutdown mid-run are cluster-smoke-only and currently unexercised (promote the `tests/conftest.py` TODO into the visible gaps list).
- Seeding GPU-path determinism is now covered by the new smoke (remove from gaps once landed).

## Not doing
- `conftest._teardown_singletons` reaching into private manager attrs — low priority; optional public teardown helper later.

zachtheyek:
Closes #248

Test-coverage hardening: the cheap, high-value improvements from the suite audit. No behavior changes to `src/` — this is tests, CI config, and docs only.

## What changed

- **CI marker leak-guard** (`.github/workflows/tests.yml`): the run selection is now `pytest -m "not gpu and not cluster and not integration" -q`, so a test marked only `integration` (which skips the singleton-isolation fixture) can never leak into CI regardless of co-marking.
- **`Logger.stop()` regression test** (`tests/unit/test_logger.py`): builds a real `Logger`, spies on its `log_queue` via `patch.object`, calls `stop()`, and asserts **both** `log_queue.close()` and `log_queue.cancel_join_thread()` fire — converting the #221 interpreter-hang from a >90s suite timeout into a fast, localized failure.
- **`_grouped_legend_entries` unit test** (`tests/unit/test_monitor.py`): parametrized over GPU count vs `max_rows`, asserting `ncol` and the column-major group padding (invisible `Line2D` handles + empty labels) for no-wrap, no-wrap-padded, the docstring's 5|5 example, wrap, and wrap-padded cases (#217).
- **Benchmark-report smoke assertions** (`tests/integration/test_train_smoke.py`, `test_inference_smoke.py`): after the config-snapshot assert, `assert os.path.exists(.../plots/benchmark_report_{tag}.png)`. Verified both the train path (`main.py:393`) and the inference path (`main.py:891`) call `_post_benchmark_report`, and that the render target is `output_path/plots/benchmark_report_{tag}.png` (`main.py:273`).
- **`smoke_model_tag` fixture** (`tests/integration/conftest.py`): `os.environ.get("AETHERSCAN_SMOKE_MODEL_TAG", "test_v17")`, consumed by `test_inference_smoke.py` and `test_model_behavior.py` in place of the hardcoded `_MODEL_TAG`.
- **Seeding-determinism cluster smoke** (`tests/integration/test_seeding_determinism.py`, marked `integration,gpu,cluster`): under a fixed seed + `enable_op_determinism()`, builds the beta-VAE twice via `create_beta_vae_model()` and takes one real train step over identical synthetic snippets, asserting byte-identical encoder `get_weights()` — the only coverage of #207's `--tf-deterministic-ops` promise.
- **`docs/TESTING.md`**: two accepted-gap bullets (slack_handler.py observability-only; mid-run resume + clean SIGTERM/SIGINT shutdown cluster-smoke-only).

## Verified locally (TF 2.17 venv, Python 3.12)

- `pytest tests/unit/test_logger.py tests/unit/test_monitor.py -m "not gpu and not cluster" -q` → **32 passed in ~5s** (6 new: 1 logger + 5 parametrized monitor). No hang.
- All touched files AST-parse; `pytest tests/integration --collect-only` collects all 4 tests; running the integration dir yields **4 clean skips** at the `nvidia-smi` guard (not fixture errors), confirming the `smoke_model_tag` fixture wiring resolves for both consumers and the new seeding test.
- The new CI expression correctly **deselects all 4 integration tests** (`no tests collected (4 deselected)`).
- `pre-commit run --files <changed>` → all hooks pass (ruff lint + format, whitespace/eof, gitleaks).

## Deferred to CI / cluster (NOT run here)

- The three `gpu`+`cluster` integration tests cannot run without physical GPUs and cluster-resident data/models — they were **import-/collect-checked and skip-verified only**, not executed.

## Notes for the reviewer

- **`test_logger` teardown**: the new test calls `stop()` once and then `Logger._reset()` in the `finally` (not `shutdown_logger()`), because `stop()` is **not idempotent** — `QueueListener.stop()` raises `AttributeError` on a second call (verified empirically), so `shutdown_logger()` after an explicit `stop()` would crash the teardown. Same isolation guarantee, avoids the double-stop. It also disposes the real queue in the `finally` so the mocked-out disposal can't leave a live feeder thread (the very hang #221 fixed).
- **Seeding smoke — API names verified against source**, but flagged for a human with cluster access to double-check by actually running it:
  - `create_beta_vae_model()` (`aetherscan.models`), `BetaVAE.compute_total_loss(main_data, true_data, false_data, target_data, training=True)`, `vae.trainable_variables`, `vae.optimizer.apply_gradients(...)`, `vae.encoder.get_weights()`, `tf.clip_by_global_norm(grads, 1.0)` — all mirror `train.py`'s `_distributed_train_step` + `_apply_gradients`.
  - Seed/determinism setup mirrors `train.py` (`tf.random.set_seed(seed)` before construction, inside `strategy.scope()`; `tf.config.experimental.enable_op_determinism()`).
  - The test bootstraps the config singleton with `init_config()` (integration tests skip the autouse config-init fixture) and pins `get_config().training.seed`, since `create_beta_vae_model()` reads `get_config()`.
  - **The core assertion — that byte-identical weights actually hold on real GPU hardware — is unverifiable locally by nature; that is the behavior the test exists to pin.**
  - Deliberate minimal choice: the test generates fixed synthetic snippets in-process and therefore does **not** take `cluster_paths` (no cluster data is read). It keeps the `gpu`+`cluster` markers so it only runs alongside the other GPU-host smokes. Add `cluster_paths` if you prefer an exact structural mirror.
- **`docs/TESTING.md` overlaps with open PR #247** (which edits the test-tree and logger bullets). These edits live only in the deliberate-gaps prose to avoid conflict, but a trivial rebase may be needed depending on merge order. No existing seeding-determinism gap bullet existed to remove.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -59,4 +59,4 @@ jobs:
       # GPU- and cluster-marked tests need physical GPUs / cluster-resident data;
       # everything else (including slow CPU TF graph tests) runs here
       - name: Run test suite
-        run: pytest -m "not gpu and not cluster" -q
+        run: pytest -m "not gpu and not cluster and not integration" -q
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -138,6 +138,17 @@ ever observed on real training runs. Gate 2 — the SNR→confidence monotonicit
 training pipeline's own injection code and scores them in-process against the persisted
 VAE+RF, so it is `gpu`+`cluster`-only by nature (it can never run in CI).
 
+Two further gaps are accepted as-is (neither affects pipeline correctness):
+
+- **`slack_handler.py`** carries no unit tests — it is observability-only (a `logging.Handler`
+  that posts records to Slack), so its absence never affects pipeline results. Its network-free
+  helpers — the consecutive-failure cooldown, exponential backoff, and level-to-color coding —
+  would be cheap to unit-test in isolation if the module is ever hardened.
+- **End-to-end training-loop resume and clean SIGTERM/SIGINT shutdown mid-run** are exercised
+  only by real cluster runs: nothing today asserts that a mid-run interrupt tears down resources
+  cleanly and resumes from the persisted checkpoint/manifest. This promotes the standing
+  [`tests/conftest.py`](../tests/conftest.py) TODO into the visible gaps list.
+
 ## Markers
 
 | Marker | Meaning | In default selection? |
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -30,6 +30,14 @@ def cluster_paths():
     )
 
 
+@pytest.fixture
+def smoke_model_tag():
+    """The persisted model tag the consumer smokes (inference + model-behavior) score against,
+    overridable via AETHERSCAN_SMOKE_MODEL_TAG. Defaults to the blpc3 dummy model (test_v17);
+    point it at whatever tag test_train_smoke just produced on a fresh cluster."""
+    return os.environ.get("AETHERSCAN_SMOKE_MODEL_TAG", "test_v17")
+
+
 # NOTE: come back to this later
 @pytest.fixture
 def run_pipeline():
--- a/tests/integration/test_inference_smoke.py
+++ b/tests/integration/test_inference_smoke.py
@@ -28,22 +28,21 @@ import pytest
 pytestmark = [pytest.mark.integration, pytest.mark.gpu, pytest.mark.cluster]
 
 # NOTE: this smoke is coupled to a specific dummy model that must exist on the cluster; the
-# skip guard below handles its absence gracefully. Future improvement for a contributor who
-# wants to decouple it: make the tag configurable via an env var (e.g.
-# AETHERSCAN_SMOKE_MODEL_TAG) or a pytest option/marker so the smoke isn't wired to one artifact.
-_MODEL_TAG = "test_v17"  # persisted dummy model on blpc3
+# skip guard below handles its absence gracefully. The tag is overridable via the shared
+# smoke_model_tag fixture (AETHERSCAN_SMOKE_MODEL_TAG, default test_v17) so this smoke can point
+# at whatever tag test_train_smoke just produced instead of being wired to one fixed artifact.
 _CSV_NAME = "subset_test.csv"  # 2 complete 6-observation cadences
 
 
-def test_inference_smoke(cluster_paths, run_pipeline):
+def test_inference_smoke(cluster_paths, run_pipeline, smoke_model_tag):
     if shutil.which("nvidia-smi") is None:
         pytest.skip("requires a GPU host (nvidia-smi not found)")
 
     data_path, model_path, output_path = cluster_paths
 
-    encoder = os.path.join(model_path, f"vae_encoder_{_MODEL_TAG}.keras")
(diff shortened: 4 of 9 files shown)


zachtheyek:
Addressed the one nit (`enable_op_determinism` process-global side effect) with the suggested docstring flag. Everything else in the review was a confirm-and-ship — thanks for cross-checking the seeding-determinism API mirror + the `_grouped_legend_entries` math against source.

Note: `docs/TESTING.md` here edits the deliberate-gaps prose (lines ~133+), disjoint from #247's test-tree/coverage-bullet edits (lines 47–88); a trivial rebase over #247 composes both.

@claude please re-review.

PR #253, opened 2026-07-23
feat(benchmarks): GPU benchmark for Beta-VAE train step + encoder inference

### LINKED ISSUE #252: Add GPU benchmarks for the Beta-VAE (training step + encoder inference)
## Context / discussion

Raised during the v1.0.0 benchmarking pass: the `benchmarks/` suite times only
hot **CPU** kernels (`bench_normality`, `bench_injection`, `bench_lognorm_downsample`,
`bench_pfb_vs_spline`), so the pipeline's **GPU** work — the Beta-VAE training step and encoder
inference — is entirely unmeasured. We should be able to benchmark every stage of the pipeline,
including the GPU-resident ones.

## Gap

There is no way to measure, per GPU:

- Beta-VAE **training-step** throughput and peak VRAM
- **encoder inference** throughput and peak VRAM
- the **largest per-replica batch that fits** a given card's VRAM — blpc3's 96 GB Blackwell and
  bla0's 16 GB A4000 differ by 6×, and nothing in the suite tells you how to size the per-replica
  batch to the hardware.

## Proposal

Add `benchmarks/bench_gpu.py` (GPU-only, runs in the NGC container). It builds the real model via
`create_beta_vae_model()` on a single GPU and drives synthetic batches of the true pipeline shapes
(train: a cadence `(B, 6, 16, 512)`, so one step is `18*B` encoder + `6*B` decoder passes, matching
`_distributed_train_step`; encode: single observations `(B, 16, 512, 1)`), reports throughput and
peak VRAM, and `--find-max` sweeps the per-replica batch to the largest power of two that fits.

## Initial findings (blpc3, 1× RTX PRO 6000 Blackwell, 96 GB)

Training step:

| per-replica batch | cadences/s | peak VRAM |
|---|---|---|
| 128 | 2,986 | 2.81 GB |
| 256 | 2,892 | 5.21 GB |
| 512 | 2,852 | 10.15 GB |
| 1024 | 2,826 | 19.79 GB |
| 2048 | 2,786 | 38.76 GB |
| 4096 | 2,742 | 76.52 GB |
| 8192 | OOM | — |

Encoder inference:

| per-replica batch | obs/s | peak VRAM |
|---|---|---|
| 512 | 198,963 | 0.52 GB |
| 1024 | 231,065 | 1.01 GB |
| 2048 | 197,361 | 1.72 GB |
| 4096 | 192,104 | 3.39 GB |

**The Beta-VAE is compute-bound, not VRAM-bound.** Training throughput peaks at per-replica batch
~128 (2.81 GB) and does not improve — it slightly declines — as the batch grows to the 76 GB VRAM
ceiling; inference peaks near batch ~1024 using under 2 GB. On a large-VRAM card the spare memory
cannot be converted to throughput for this compact model, so the defaults (train 128, inference
2048) are already near-optimal.

## Related observations (noted for the record; separate follow-ups)

- The production **default config cannot run on 5 GPUs** (blpc3): `effective_batch_size 3072` is not
  divisible by `per_replica_batch_size 128 × 5 = 640`, and `per_replica_val_batch_size 80 × 5 = 400`
  divides none of `val_samples 99840`, `num_samples_rf 99840`, or `latent 240×4=960` (all divide 320
  for 4 GPUs and 480 for 6 GPUs, but not 400). Worked around for benchmarking with
  `--effective-batch-size 2560 --per-replica-val-batch-size 64`.
- `utils/find_optimal_configs.py`'s solver **fails to auto-solve** that 5-GPU case ("No valid
  configuration found within search ranges") even though a valid config exists — its search ranges
  likely need widening.
- A single encoder forward whose conv feature maps exceed 2^31 elements (batch ≳ 8192) trips an
  **uncatchable TensorFlow int32 launch-config abort** rather than a clean OOM; `bench_gpu.py` caps
  `--find-max` at 4096 to stay clear of it.

PR to follow.

zachtheyek:
Closes #252

Adds `benchmarks/bench_gpu.py` — the first GPU benchmark in the suite (the existing scripts time
only hot CPU kernels). It builds the real Beta-VAE (`create_beta_vae_model`) on a single GPU and
drives synthetic batches of the true pipeline shapes to measure per-GPU throughput and peak VRAM,
sweeping the per-replica batch to find the largest that fits.

## What it measures

- **`--mode train`** — a full training step: `compute_total_loss` (reconstruction + KL + both
  clustering heads) → `tape.gradient` → global-norm clip (1.0, matching `_apply_gradients`) → Adam
  apply. A training example is a cadence `(B, 6, 16, 512)`, so one step runs `18*B` encoder + `6*B`
  decoder passes — exactly like `_distributed_train_step`.
- **`--mode encode`** — encoder forward on observations `(B, 16, 512, 1)`, i.e. the inference path.
- **`--find-max`** doubles the per-replica batch until OOM and reports the largest power of two that
  fits; `--batch-sizes a,b,c` measures explicit sizes.

Peak VRAM is per GPU (each replica holds a full model copy → combined ≈ ×replicas). Single-GPU +
memory growth (mirrors `main.py`) so the VRAM figures track real usage; model built inside a
`OneDeviceStrategy` scope per the repo's strategy-scope rule.

## Key finding

The Beta-VAE is **compute-bound, not VRAM-bound**. On blpc3 (RTX PRO 6000, 96 GB) training
throughput peaks at per-replica batch ~128 (2.81 GB) and does *not* improve toward the 76 GB VRAM
ceiling; inference peaks near batch ~1024 using <2 GB. The defaults (train 128, inference 2048) are
already near-optimal — a large-VRAM card's spare memory can't be converted into throughput for this
compact model. Full baseline tables are in `benchmarks/README.md`.

## Notes / caveats

- **GPU-only**, runs in the NGC container — like the `gpu`/`cluster` integration tests, it is not
  part of the CI CPU surface, and (consistent with the deliberately framework-free `benchmarks/`
  suite) it is not pytest-collected; it prints a table and writes a JSON result.
- `--find-max` is capped at `--max-batch 4096`: an encoder forward whose conv feature maps exceed
  2^31 elements (batch ≳ 8192) trips an **uncatchable** TF int32 launch-config abort rather than a
  clean OOM. 4096 covers the training VRAM ceiling and a generous inference range.

## Verification

- `ruff check` / `ruff format` clean; pre-commit clean.
- Ran end-to-end on blpc3 in the container: `--mode train --find-max` (max batch 4096 @ 76.5 GB) and
  `--mode encode --batch-sizes 256..4096`; numbers are the tables in the README.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -147,6 +147,32 @@ large-VRAM card (e.g. the 96 GB Blackwell) the spare memory cannot be converted 
 this compact model, so the defaults (train 128, inference 2048) are already near-optimal — size the
 per-replica batch for constraint-divisibility and convenience, not to fill VRAM.
 
-The numbers above are single-GPU on blpc3. Multi-GPU scaling (`--num-gpus`), the
-`--accumulation-steps` cadence, and the bla0 (6× A4000 16 GB) baselines are captured on the
-release-benchmark runs and land with the System-Requirements update (#183).
+### Baseline: bla0 (6× RTX A4000 16 GB, NGC 25.02, July 2026)
+
+bla0 is the release-training host. Single-GPU training step and encoder inference:
+
+| per-replica batch | train cadences/s | train VRAM | encode obs/s | encode VRAM |
+|---|---|---|---|---|
+| 128 | 795 | 2.76 GB | 47,868 | 0.12 GB |
+| 256 | 819 | 5.26 GB | 50,364 | 0.19 GB |
+| 512 | 820 | 9.99 GB | 51,016 | 0.36 GB |
+| 1024 | OOM | — | 52,462 | 0.61 GB |
+| 2048 | — | — | 53,077 | 1.18 GB |
+| 4096 | — | — | 53,001 | 2.32 GB |
+
+Multi-GPU (all 6 A4000s, per-replica batch 128, **aggregate** throughput):
+
+| config | throughput | scaling | VRAM/GPU |
+|---|---|---|---|
+| train, 1 GPU | 795 cad/s | 1.0× | 2.76 GB |
+| train, 6 GPUs | 3,087 cad/s | 3.9× | 3.21 GB |
+| train, 6 GPUs, `--accumulation-steps 4` | 4,272 cad/s | — | 9.91 GB |
+| encode, 6 GPUs (batch 1024) | 214,451 obs/s | 4.1× | 0.78 GB |
+
+The A4000 is ~3.8× slower per GPU than the Blackwell (795 vs 2,986 cad/s at batch 128), consistent
+with its lower compute; the model is still compute-bound (throughput flat as batch grows toward the
+16 GB ceiling — training OOMs above batch 512). MirroredStrategy scales ~3.9–4.1× across the 6
+replicas (all-reduce overhead eats the rest). `--accumulation-steps 4` raises throughput (4,272 vs
+3,087 cad/s — one optimizer apply per 4 micro-batches instead of every step) at the cost of the
+persistent gradient accumulator's VRAM (3.21 → 9.91 GB/GPU). At the default batch 128, training uses
+~2.8 GB/GPU — comfortably within the 16 GB budget.
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -18,6 +18,10 @@ python benchmarks/bench_pfb_vs_spline.py        # PFB static equalization vs per
 ./utils/run_container.sh python benchmarks/bench_normality.py
 ```
 
+`bench_gpu.py` is a different animal — it profiles the Beta-VAE on a real GPU (throughput +
+peak VRAM, with a batch-size sweep) rather than a CPU kernel, so it only runs inside the
+container on a cluster. See [GPU benchmark](#gpu-benchmark) below.
+
 Common flags: `--repeats N` (default 3; ops/s is reported from the best repeat) and
 `--output PATH` for the JSON result. Each script also exposes size knobs (`--width`,
 `--injections`, `--cadences`) — the defaults match production shapes, so stick to them when
@@ -31,6 +35,7 @@ comparing against the baselines.
 | `bench_injection.py` | `data_generation.new_cadence` (setigen narrowband injection into a stacked 96x4096 cadence) | Training data generation (`train.round_XX.data_generation`) |
 | `bench_lognorm_downsample.py` | per-observation `downscale_local_mean` (x8) and per-cadence `log_norm` | Stamp extraction downsample + inference load (`inference.*.load_lognorm`) |
 | `bench_pfb_vs_spline.py` | `pfb.equalize_passband` (static response divide) vs `preprocessing._spline_flatten_bandpass` (order-16 fit) on one 1M-bin coarse channel | Bandpass flattening inside energy detection |
+| `bench_gpu.py` | Beta-VAE training step (`compute_total_loss` + gradients + clipped Adam) and encoder forward on one GPU | VAE training (`train.round_XX`) and encoder inference — **GPU-only, see below** |
 
 ## Baseline numbers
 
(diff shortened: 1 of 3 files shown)


zachtheyek:
This branch evolved substantially since the first pass, and is now **validated on-cluster** (bla0, 6× A4000). Current state (`0a8ea75`..`3d5c7b3`):

- **Multi-GPU `bench_gpu.py`** (`--num-gpus`, MirroredStrategy) — resolves all 6 first-pass nits (mutually-exclusive `--find-max`/`--batch-sizes`, `example_shape` in JSON, `--warmup` guard, module-level RNG, SI-GB label, argparse hardening).
- **`--accumulation-steps`** — models `_train_epoch`'s accumulate-then-apply cadence.
- **`bench_rf.py`** — the previously-unmeasured CPU RF stage (framework-free).
- **A real multi-GPU bug, found *and fixed* by the on-cluster smoke.** The first accumulation implementation applied pre-reduced grads in cross-replica context — which core-dumps under MirroredStrategy (`Adam slot on GPU:0 accessed from GPU:1`; `apply_gradients` with a single reduced grad doesn't broadcast to the replicas' mirrored slots). Single-GPU masked it. Rewrote to the canonical **per-replica accumulator pattern**: accumulate into mirrored accumulator variables inside `strategy.run`, then apply once *inside* `strategy.run` so the all-reduce happens on apply. `_make_step_fn` is now encode-only.
- **bla0 A4000 baselines** added to `benchmarks/README.md` (single/multi-GPU + accumulation).

**On-cluster validation (bla0, all paths clean):** train single 795 / 6-GPU 3,087 (~3.9×) / accumulation K=4 4,272 cad/s @ 9.91 GB; encode single 53k / 6-GPU 214k obs/s (~4.1×). The multi-GPU + accumulation paths (the ones that couldn't be tested locally — no GPU, TF deadlocks on the dev box) are what the smoke exercised and fixed.

Fresh eyes welcome on the accumulator pattern in `_measure` in particular. @claude

zachtheyek:
Thanks — all three notes addressed in `63b54ec` (comment-only; no logic change, so the on-cluster bla0 baselines still stand as measured).

**1. Reduce/clip ordering (`bench_gpu.py:180`).** Correct catch, and I've made the claim honest rather than changing the path. Softened the module docstring from "exactly as `_train_epoch`" to *cadence*-exact, and added the precise divergence note in `_apply_and_reset`: production reduces-then-clips (`clip(mean_r(acc_r)/K)`), this clips-per-replica-then-all-reduces-on-apply (`mean_r(clip(acc_r/K))`), differing only when a replica's global norm exceeds `_CLIP_NORM`, compute-identical either way.

I deliberately **did not** adopt the explicit `all_reduce(MEAN)`-then-clip variant, for three reasons: (a) it changes the validated apply path, so I'd have to re-run the multi-GPU + accumulation smoke on bla0 to re-confirm no cross-device crash and that the numbers hold — churn for a path whose output this benchmark doesn't measure; (b) the benchmark reports throughput + peak VRAM, both of which are order-invariant here; (c) it keeps the accumulate-then-apply structure minimal. The one-line note now documents the divergence exactly as you suggested, so a reader porting this to a *correctness* harness knows where to switch.

**2. K=1 walks the accumulator path (`bench_gpu.py:155`).** Acknowledged — leaving as-is intentionally. A fused K=1 fast-path would be a *second* code path that no longer exercises the accumulate→apply structure the multi-GPU fix lives in, so the "plain apply-every-step" row would stop sharing the path it's meant to baseline. The measured overhead is in the noise (1→6 GPU is 795→3,087, 3.9×; a per-param `assign_add` + zero-out fuses into the one `@tf.function`), so the extra branch would cost more in divergence risk than it saves. If a future K=1 number looks off, that's the first place I'd add the fast-path.

**3. Primary-replica sync barrier (`bench_gpu.py:191`).** Agreed the reasoning was non-obvious — expanded the `sync` comment to state it: every train apply ends in a collective all-reduce, so a stream drained past step N on any replica implies all replicas finished that step's all-reduce, making a single mutated-`MirroredVariable` read a sufficient *cross-replica* barrier (not merely a primary-replica one). Locks it against a future "fix."

**README.** Named the mechanism behind the `--accumulation-steps` bump — one apply + clip + all-reduce per K micro-batches amortizes the fixed per-apply overhead over 4× the cadences — so it doesn't have to be reasoned out.

`bench_rf` parity, the six first-pass nits, and the accumulator-pattern correctness all as you confirmed. If this reads clean I'll stop here and leave it for maintainer merge.

@claude

PR #255, opened 2026-07-23
feat(config): defaults valid on 4/5/6 GPUs + robust find_optimal_configs solver

### LINKED ISSUE #254: Config defaults invalid on 5-GPU hosts + find_optimal_configs solver misses valid configs
## Context / discussion

Surfaced during v1.0.0 benchmarking: the production **default config cannot run on a 5-GPU host**
(blpc3, the Blackwell cluster). With `num_replicas=5`, `effective_batch_size=3072` is not divisible
by `per_replica_batch_size 128 × 5 = 640`, and `per_replica_val_batch_size 80 × 5 = 400` divides none
of the val split (99840), `num_samples_rf` (99840), or `latent_viz*4` (960). The defaults happen to be
valid on 4- and 6-GPU hosts (320 and 480 divide those counts) but not 5.

### 1. Defaults valid on every supported GPU count (4/5/6)

Set `effective_batch_size = 7680` (= `lcm(128*4, 128*5, 128*6)`, and it divides the 399360-sample
train split) and `per_replica_val_batch_size = 16` (the largest val batch whose global size divides the
val/RF/latent counts for 4/5/6 GPUs). `per_replica_batch_size` stays 128 (the throughput-optimal per
the GPU benchmark) and `num_samples` are unchanged.

Performance: per-GPU **training throughput is unchanged** (set by `per_replica_batch_size=128`;
`effective_batch_size` only changes the gradient-accumulation count). `V=16` runs validation /
latent-generation at a smaller batch (slightly launch-overhead-bound per the encode benchmark) but off
the training hot path, so the cost is negligible. `E=7680` is 2.5× the previous 3072 — a fixed-GPU-count
run can override `--effective-batch-size 3072` (valid on 4 or 6 GPUs) to keep the previous behavior.

### 2. Harden `find_optimal_configs.py`'s cross-param solver

The solver (`_solve_cross_param_constraints`) has three compounding bugs that make it return "No valid
configuration found" even when one exists:

1. `_SEARCH_RANGES['per_replica_val_batch_size'] = (256, 768)` — the real val batch is ~16–96, and any
   `global_val ≥ 1024` can't divide `latent_viz*4 = 960`. The range excludes the entire solution region.
2. With `step=1` on the batch/val ranges the grid is ~2.8e9 candidates, which exceeds the 1e7
   `max_candidates` budget, so the solver bails to `None` before searching.
3. `_check_cross_constraints` (its feasibility test) omits the `latent_viz*4 % global_val` check that
   `collect_validation_errors` enforces, so a returned "solution" can be latent-invalid.

Fix: generate small, exact candidate sets from the divisibility structure (divisors of `train_samples`
for `effective_batch_size`; divisors of `gcd(val, rf, latent)` for `global_val`) so the search is tiny
and complete, and make `_check_cross_constraints` latent-aware (closing #3, consistent with
`collect_validation_errors`). The shared checker stays the general authority — this only makes candidate
generation smart and the feasibility test correct, so it is not overfit to the current batch problem.

Both the `find_optimal_configs.py` utility and the inline "Suggested fixes" block in `validate_args`
call the same solver, so both improve.

## Scope

`src/aetherscan/config.py` (defaults), `src/aetherscan/cli.py` (solver + checker), tests, and doc
touch-ups (`docs/CONFIG_AND_CLI.md`, `docs/TRAINING_PIPELINE.md`).

zachtheyek:
Closes #254

Two coupled changes: make the default training config valid on every supported GPU count (4/5/6),
and fix the `find_optimal_configs.py` solver that couldn't find configs that exist.

## 1. Defaults valid on 4/5/6 GPUs (`config.py`)

The production defaults failed to run on a **5-GPU host** (blpc3): `effective_batch_size 3072` isn't
divisible by `per_replica_batch_size 128 × 5 = 640`, and `per_replica_val_batch_size 80 × 5 = 400`
divides none of the val split (99840), `num_samples_rf` (99840), or `latent_viz*4` (960). They
happened to be valid on 4- and 6-GPU hosts, not 5.

| field | old | new | why |
|---|---|---|---|
| `per_replica_batch_size` | 128 | **128** | throughput-optimal per the GPU sweep — unchanged |
| `effective_batch_size` | 3072 | **7680** | `lcm(128×{4,5,6})`, divides the 399360 train split → valid on 4/5/6 |
| `per_replica_val_batch_size` | 80 | **64** | global size divides val/rf/`latent_viz*4` on 4/5/6 |
| `latent_viz_num_cadences_per_type` | 240 | **960** | raised in lockstep so `latent_viz*4` stays divisible by the larger val batch |

`num_samples_*` unchanged. **Performance:** per-GPU **training throughput is unchanged** (set by
`per_replica_batch_size=128`; `effective_batch_size` only changes the gradient-accumulation count).
`effective_batch_size=7680` is 2.5× the previous 3072 — a fixed-GPU-count run can override
`--effective-batch-size 3072` (valid on 4 or 6 GPUs) to keep the old value; the maintainer will watch
loss convergence during the final training run. `per_replica_val_batch_size=64` (rather than the bare
divisibility floor of 16 at `latent_viz=240`) keeps validation / latent-generation out of the
small-batch launch-overhead regime the encode benchmark shows — a net win over a full run's ~2000
validations — at the cost of 4× denser latent-space plots and a negligible per-round viz-gen bump.

Verified valid for 4/5/6 via `find_optimal_configs.py --num-gpus 4,5,6 train`.

## 2. Robust cross-param solver (`cli.py`, `find_optimal_configs.py`)

`_solve_cross_param_constraints` returned "No valid configuration found" even when one existed, from
three compounding bugs:

1. **Miscalibrated range** — `_SEARCH_RANGES['per_replica_val_batch_size'] = (256, 768)`, but the real
   val batch is ~16–96, and any `global_val ≥ 1024` can't divide `latent_viz*4 = 960`; the range
   excluded the entire solution region.
2. **Search-space explosion** — with `step=1` the grid was ~2.8e9 candidates, over the 1e7 budget, so
   it bailed to `None` before searching.
3. **Latent constraint ignored** — `_check_cross_constraints` omitted the `latent_viz*4 % global_val`
   check that `collect_validation_errors` enforces, so a returned "solution" could be latent-invalid.

Fix: enumerate the two purely divisibility-bound fields **exactly** from the structure — the effective
batch from divisors of the train split (multiples of `per_replica_batch × nr`), the global val batch
from divisors of `gcd(val, rf, latent_total)` — so the search is tiny and complete rather than a coarse
grid that explodes or steps over the only valid values. Data sizes and per-replica batch are only
nudged (a last resort). Every candidate is confirmed by the now **latent-aware**
`_check_cross_constraints`, which stays the single source of truth (so this is not overfit to the
current batch problem — any constraint it enforces is honored). Both `find_optimal_configs.py` and the
inline `validate_args` "Suggested fixes" block use the same solver, so both improve.

## Tests

`tests/unit/test_cli_validation.py`: flipped the divisibility matrix (defaults now valid on 4/5/6);
rewrote the solver suite for the new behavior — keeps data + per-replica batch when fixing a 5-GPU
config, is latent-aware, satisfies multiple replica counts simultaneously, and the candidate budget is
a backstop. Verified locally (TF-free): all pass, and the new defaults + solver produce
`nr=5 → E=2560/V=64`, `nr=4/5/6 → E=3840/V=16`.

### DIFF
--- a/docs/CONFIG_AND_CLI.md
+++ b/docs/CONFIG_AND_CLI.md
@@ -350,8 +350,8 @@ Examples (from the script's docstring):
 # Check default config against a 4-GPU setup
 python utils/find_optimal_configs.py --num-gpus 4 train
 
-# Override --effective-batch-size and search for a fix valid on both 4 and 6 GPUs
-python utils/find_optimal_configs.py --num-gpus 4,6 train --effective-batch-size 3072
+# An override that's invalid for 5 GPUs — the solver proposes the nearest valid combination
+python utils/find_optimal_configs.py --num-gpus 5 train --effective-batch-size 3072
 
 # Check inference defaults with an invalid --overlap-fraction
 python utils/find_optimal_configs.py inference --overlap-fraction 1.5
--- a/docs/TRAINING_PIPELINE.md
+++ b/docs/TRAINING_PIPELINE.md
@@ -177,7 +177,10 @@ Each training step accumulates `accumulation_steps = effective_batch_size /
 (per_replica_batch_size × num_replicas)` micro-batch gradients before applying
 (`_train_epoch()` → `_distributed_train_step()` → `_apply_gradients()`), giving an effective
 batch of 7680 at defaults (chosen so it divides evenly on 4-, 5-, or 6-GPU hosts) regardless of
-per-GPU memory. Guards along the way: all-None
+per-GPU memory. **Note:** 7680 is ~2.5× the previous 3072, so there are ~2.5× fewer optimizer
+updates per epoch; LR-schedule behavior calibrated to the old cadence may differ. On a fixed 4- or
+6-GPU host you can pass `--effective-batch-size 3072` to restore the old cadence (it stays valid
+there). Guards along the way: all-None
 gradient micro-batches are skipped, accumulated gradients are averaged over successful
 micro-steps, NaN/Inf gradients raise immediately, and the global gradient norm is clipped at
 1.0 with the pre-clip norm recorded per step (that's the `clipping_rate` statistic).
--- a/docs/TRAINING_PIPELINE.md
+++ b/docs/TRAINING_PIPELINE.md
@@ -42,7 +42,7 @@ as done, so re-running the identical command resumes exactly where the last atte
    in the producer process while round *k* trains.
 3. **Build datasets** — `prepare_distributed_train_dataset()` over the round's memmaps
    (stratified 80/20 train/val split on the labels array).
-4. **Prepare the latent-viz batch** (first round only) — 240 held-out val cadences per signal
+4. **Prepare the latent-viz batch** (first round only) — 960 held-out val cadences per signal
    type, persisted across rounds so latent-space snapshots aren't confounded by curriculum
    distribution shift.
 5. **Epoch loop** — `_train_epoch()` + `_validate_epoch()`, ~21 `training_stats` rows per
@@ -176,7 +176,8 @@ of `effective_batch_size` (train) and the global val batch size. With `shuffle=F
 Each training step accumulates `accumulation_steps = effective_batch_size /
 (per_replica_batch_size × num_replicas)` micro-batch gradients before applying
 (`_train_epoch()` → `_distributed_train_step()` → `_apply_gradients()`), giving an effective
-batch of 3072 at defaults regardless of per-GPU memory. Guards along the way: all-None
+batch of 7680 at defaults (chosen so it divides evenly on 4-, 5-, or 6-GPU hosts) regardless of
+per-GPU memory. Guards along the way: all-None
 gradient micro-batches are skipped, accumulated gradients are averaged over successful
 micro-steps, NaN/Inf gradients raise immediately, and the global gradient norm is clipped at
 1.0 with the pre-clip norm recorded per step (that's the `clipping_rate` statistic).
--- a/src/aetherscan/cli.py
+++ b/src/aetherscan/cli.py
@@ -2391,11 +2391,6 @@ def collect_validation_errors(
 # gymnastics, and so the validation + proposal logic stays colocated.
 # ---------------------------------------------------------------------------
 
-# Grid-search ranges for the six interdependent training params that govern
-# how data is divided across replicas, batches, and train/val splits. Most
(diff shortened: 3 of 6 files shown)


zachtheyek:
Thanks — positive review, no blockers. Addressed the nits in `3db25be`:

1. **Stale comment (`cli.py:2394`)** — fixed. Removed the pre-refactor "grid-searches these ranges" block that contradicted the new one right below it.
2. **Solver docstring wording (`cli.py:2500`)** — fixed. Softened to say the data sizes + per-replica batch are *seeded with the base value* (so preferred when a solution keeps them) but can still move when L1-closer — with the explicit `4/5/6 → p=64` example you called out.
4. **Strict L1 test (`test_cli_validation.py`)** — added `test_solver_picks_the_l1_nearest_not_the_first_valid`, which asserts the *exact* nearest values (`E=2560`, not merely "some valid E"; val batch 16 from 80), so a refactor returning the first valid candidate instead of the closest now fails. On the "no genuinely-unsolvable base" gap: the `None` path is covered by the budget-zero test, and a *structurally* unsolvable base is hard to construct meaningfully — the solver can nudge `per_replica_batch`/data sizes across their neighborhoods to satisfy almost any input (which is the intended behavior), so I left that as-is rather than adding a contrived case. Happy to add one if you can point at a realistic unsolvable input.
5. **Convergence-cadence + override discoverability (`config.py:210`)** — documented in `TRAINING_PIPELINE.md` (the training runbook) right where the 7680 effective batch is explained: the ~2.5× fewer updates/epoch caveat + the `--effective-batch-size 3072` restore-the-old-cadence escape hatch for 4/6-GPU hosts. Put it there rather than `KNOWN_ISSUES.md` since it's documented config behavior, not a bug.

**Not taking one (rebut):**
3. **Reason string on `None` return (`cli.py:2581`)** — leaving as-is for now. It's a nice-to-have, but it changes the return contract (`dict | None` → carry a reason) and touches both callers (`find_optimal_configs.py` + the inline `validate_args` block) for an ops-triage convenience that only matters when `max_candidates` is set low — which no production caller does (the default 5e6 backstop is never hit by the tiny divisor-generated candidate sets). If solver-`None` ever shows up in the wild I'll add the reason then.

@claude

PR #259, opened 2026-07-23
feat(train): parallelize RF SHAP explainability across CPU cores

### LINKED ISSUE #258: feat(train): parallelize RF SHAP explainability across CPU cores (the ~80h single-threaded tail)
## Problem

The RF-explainability step at the tail of training (`train.py::_compute_or_load_shap_values`) runs `shap.TreeExplainer` **single-threaded**: shap's C TreeSHAP extension has no OpenMP and no `n_jobs`, and the RF's own `n_jobs` doesn't apply (shap re-walks the trees itself). On a production RF (1000 trees, unbounded depth) this makes the SHAP tail pathologically slow — it runs once per training run, at the very end, and stalls a multi-day job for *days*.

Measured on the real `test_v50` RF (1000 trees, max_depth 53), single-threaded:

| pass | per-sample | production count | single-thread |
|---|---:|---|---:|
| summary | 2.16 s | 5000 | ~3.0 h |
| **interaction** | **183 s** | **1500** | **~76 h** |
| log-loss | (cheaper) | 5000×1000bg | (secondary) |

→ the full tail is **~80 h, ~95% interaction**. (The historically-observed "~2.5 h" was mid-computation, not completion.)

## Root cause / opportunity

SHAP values are **per-sample independent** (embarrassingly parallel), but stock shap doesn't parallelize. Chunking samples across processes — each calling stock `shap.TreeExplainer` on its chunk — is not a fork of the algorithm, just parallelism.

## Solution (this issue)

A **process-pool SHAP wrapper** for all three passes (summary, interaction, log-loss):
- **Uses all available cores by default** (`config.manager.n_processes`, already `cpu_count()`).
- **Rebuilds a stock `TreeExplainer` inside each worker** (never pickles a pre-built explainer — that segfaults large-model workers, shap #1204).
- Spawn workers in a **TF-free module** so they don't drag TensorFlow into each process; **pin BLAS/OpenMP threads to 1 per worker** to avoid oversubscription.
- **Byte-identical to the single-threaded result** (verified `allclose`).

Measured **~40–45× on a 96-core node** (RAM-cheap, ~1.4 GB/worker → core-bound). On a shallow RF this takes the tail to **~6 min**.

## Why not GPU

`GPUTreeExplainer` is faster on interaction (~5 s/pass), but the experiment found it **not worth the maintenance** for a few minutes saved at the end of a multi-day/-week run:
- Requires a **from-source CUDA build of an experimental shap path** baked into the container/conda (the stock wheel has no `_cext_gpu`).
- **Hard `< 32` constraint** (a fixed CUDA warp-lane / path-length overflow, `rapidsai/gputreeshap`) — precisely, ≤ 31 *distinct features per root-to-leaf path* — so it needs `max_depth ≤ 31`.
- Its **interventional log-loss path is broken**: `_cext_gpu.cu` drops the output-transform pointer and silently returns raw-margin numbers (shap #4270/#3936/#1726, unfixed in 0.46.0), failing the additivity axiom.

The full comparison + a "how to switch to GPU later" runbook + the depth constraint land in `docs/TRAINING_PIPELINE.md`. We keep `max_depth` unbounded; the constraint is documented only as GPU context.

## Acceptance criteria

- [ ] MP wrapper drives summary + interaction + log-loss across `manager.n_processes` workers by default.
- [ ] Output byte-identical (`np.allclose`) to the single-threaded path; unit test proving it.
- [ ] TF-free worker module; threads pinned; pool cleaned up.
- [ ] `docs/TRAINING_PIPELINE.md` notes the MP decision + GPU findings + switchover + the depth/distinct-features constraint.

zachtheyek:
## Summary

Closes #258. `train.py::_compute_or_load_shap_values` ran `shap.TreeExplainer` **single-threaded** — shap's C TreeSHAP has no OpenMP / `n_jobs`, and the RF's own `n_jobs` doesn't apply (shap re-walks the trees itself) — so on a production 1000-tree forest the **interaction pass alone took ~76 h** (measured ~183 s/sample × 1500), stalling the end of every training run for days. This parallelizes it across all CPU cores.

## What changed

- **`aetherscan/shap_parallel.py`** (new, TF-free): a process-pool wrapper. Chunks samples across `manager.n_processes` (= `cpu_count()` by default) workers, each rebuilding a **stock** `TreeExplainer` on its chunk — rebuild-per-worker avoids the shap #1204 segfault from pickling a pre-built explainer; workers load the RF from its persisted joblib; BLAS threads pinned. Byte-identical to the serial result; **~40–45× on a 96-core node**. Runs under a **`forkserver` context with an empty preload** so workers stay lightweight: `train.py` imports TF at module level, and `spawn` would re-import the parent's `__main__` (`aetherscan.main` → TF → the whole training stack) into every worker — the fork server is spawned once, clean, and workers fork from it importing only `shap_parallel` + `shap`. (This was corrected from an initial `spawn` in a follow-up commit during self-review.)
- **`train.py`**: `_compute_or_load_shap_values` drives summary + interaction + log-loss through `parallel_shap`; `select_positive_class_shap` moved to `shap_parallel` (shared with the workers). Expected-value uses one lightweight in-process explainer.
- **Tests**: `tests/unit/test_shap_parallel.py` (new, TF-free) — the shape-normalization tests (relocated from `test_train_utils.py`) plus a parametrized MP-vs-single-process `allclose` correctness test for all three passes.
- **`docs/TRAINING_PIPELINE.md`**: a SHAP-performance subsection — why CPU MP, the GPU findings (GPUTreeExplainer is faster on interaction but needs a from-source CUDA build + a `< 32` **distinct-features-per-path** warp-lane constraint + a broken interventional log-loss path, shap #4270/#3936/#1726), and a switchover runbook. `max_depth` stays **unbounded**; the constraint is documented only as GPU context.

## Validation

- Byte-identity is enforced by the new unit test (single vs 4-worker `allclose`, all three passes) — CI runs it (shap is a runtime dep).
- Empirically ~40–45× on bla0 (96 cores) during the experiment behind #258; RAM-cheap (~1.4 GB/worker → core-bound).
- The `.nosync` TF-import deadlock blocks the full pytest suite locally, so the shape-normalization + `ValueError` paths were checked standalone; the MP path is covered by CI.

## Notes

- The rewritten block drops the resolved `# NOTE: come back to this later` questions (now answered in the docs).
- `max_depth` intentionally unchanged (unbounded, sklearn default).

### DIFF
--- a/docs/TRAINING_PIPELINE.md
+++ b/docs/TRAINING_PIPELINE.md
@@ -403,6 +403,58 @@ share `rf_shap_values_{tag}.joblib` (computed once, cached).
 | `rf_oob_accuracy_curve_{tag}.png` (from `plot_rf_ensemble_accuracy_curve`) | Cumulative accuracy vs number of trees (val + train-subsample baseline), elbow annotated. Also persists the per-tree `ensemble_val_accuracy` series to `training_stats` (`model_name='rf'`, `epoch_number` = tree count) for the dashboard RF tab — so the DB series only lands when `rf_plots` succeeds. | Should saturate well before 1000 trees; if it's still climbing at the end, raise `rf.n_estimators`. |
 | `rf_latent_decision_boundary_nn{n}_md{m}_{tag}.png` | RF P(true) contour over each persisted cadence-level UMAP plane, val points + 0.5 contour. | A coherent boundary separating the true classes; ragged islands = the forest partitioning noise. Depends on the UMAPs from `plot_latent_space_gif`, so `vae_plots` must have succeeded. |
 
+### SHAP explainability performance (CPU multiprocessing; GPU is a documented alternative)
+
+The five SHAP figures share `rf_shap_values_{tag}.joblib`, computed once by
+`_compute_or_load_shap_values`. shap's TreeSHAP C extension is **single-threaded** (no OpenMP, no
+`n_jobs`; the RF's own `n_jobs` does not apply — shap re-walks the trees itself), so on a 1000-tree
+forest the step is dominated by the **interaction** pass and runs for hours-to-days if left serial
+(measured ~183 s/sample on a depth-53 RF → ~76 h for 1500 interaction samples; the whole tail is
+~95% interaction).
+
+SHAP values are per-sample independent, so we **chunk the samples across all cores**
+(`aetherscan.shap_parallel`, driven by `manager.n_processes` = `cpu_count()` by default): each worker
+rebuilds a *stock* `TreeExplainer` and explains its chunk, and the results are byte-identical to the
+serial computation (measured ~40-45x on a 96-core node). This is the shipped path for all three
+passes (summary, interaction, log-loss).
+
+#### GPU is faster on interaction, but we don't use it — here's why, and how to switch
+
+shap's `GPUTreeExplainer` (GPUTreeShap) runs each pass in ~seconds regardless of sample count (~1000x
+on interaction). We benchmarked it on both clusters; it works and is **correct** for the summary and
+interaction passes (`np.allclose` vs CPU). We still ship CPU multiprocessing, because the GPU route
+buys only a few minutes at the very end of a multi-day/-week run in exchange for maintaining an extra
+CUDA + from-source dependency on an experimental shap code path:
+
+- **Not in the stock wheel.** `GPUTreeExplainer` needs a `_cext_gpu` CUDA extension that exists only
+  if shap is built from source with `SHAP_ENABLE_CUDA=1` + a CUDA toolkit — a bespoke build baked into
+  the container/conda and re-verified on every container/shap/CUDA bump.
+- **Hard depth limit (fixed lane overflow).** GPUTreeShap maps one path element per CUDA warp *lane*
+  (`warpSize == 32`, fixed on every NVIDIA arch), so an over-long root-to-leaf path overflows and the
+  kernel aborts (`Tree depth must be < 32`, core dump). Precisely, it caps the number of **distinct
+  features per root-to-leaf path at ≤ 31** (paths are de-duplicated by feature before the length
+  check), so `max_depth ≤ 31` is a *sufficient*, guaranteed-safe setting; a deeper forest can still
+  run if no path uses > 31 features, but that is not guaranteed. It is **not** fixable by a different
+  build.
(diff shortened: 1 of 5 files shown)


zachtheyek:
Thanks — all four addressed in `1bf2c01`:

1. **Always-dump the RF** (`train.py`). Agreed: the `if not os.path.exists` guard turned a stale-file case into a silent no-op, while `expected_value` uses the in-process model — a latent divergence. Now dumps `self.rf_model.model` unconditionally, so workers load exactly the model the base value is computed from. The extra ~few-hundred-MB write is dwarfed by the SHAP work, as you noted.

2. **Empty-input guard** (`shap_parallel.py`). Added `raise ValueError("cannot explain zero samples")`. I went with raising rather than fabricating an empty result, because the suggested `np.empty_like(x, shape=(0, *x.shape[1:]))` would be the wrong rank for the interaction pass (output is `(0, F, F)`, not `(0, F)`). Not reachable via the caller today, but it makes the module safe to reuse.

3. **Redundant sort** (`shap_parallel.py`). Dropped it — and removed the now-dead `chunk_index` from the worker's return, since relying on `ProcessPoolExecutor.map` preserving input order is exactly the intended invariant. Left a one-line comment stating it.

4. **"byte-identical" vs the test.** You're right the claim outran the test. But the claim is *true* — TreeSHAP is per-sample independent with no cross-sample reduction, so chunking is bitwise-identical, not merely close. So rather than soften the docstring I **tightened the test** to `assert_array_equal` (was `allclose(atol=1e-8)`), making the stronger claim test-backed.

CI was green on the prior push; this commit is lint-clean locally (the full suite can't run on the `.nosync` box — TF import deadlock — so the MP correctness path is covered by CI). Re-tagging for a second pass.

@claude

zachtheyek:
Second-pass check: all four first-pass notes are resolved in `1bf2c01` (always-dump RF, empty-input guard, dropped the redundant sort + dead `chunk_index`, tightened the correctness test to `assert_array_equal`), CI is green across 3.10/3.11/3.12 + CodeQL + pre-commit, and the branch is rebased on `master`. The prior re-tag did not surface a review — requesting a fresh look at `1bf2c01`, particularly the forkserver-with-empty-preload worker setup in `shap_parallel.py` and the always-dump change in `train.py`. If it reads clean, an LGTM is all I need.

@claude

Issue #142, opened 2026-07-17
--load-tag round_XX without --load-dir checkpoints silently resumes from a stale tag and reports success


zachtheyek:
## Summary

The natural command for resuming training after a crash — `--load-tag round_XX` **without** `--load-dir checkpoints` — **silently loads a stale, unrelated tag's final model, skips Random Forest training, and still reports "Training completed successfully!"**. No error, no non-zero exit; the run produces a model that was never trained on the intended checkpoint. Found during the #117 crash/resume verification on blpc3.

This is **pre-existing** `load_models` behavior (confirmed: `load_models` is not touched by #117's diff), but #117's disk-backed crash/resume story makes the resume path a routine operation, so the footgun is now much more likely to be hit — it is exactly the command an operator types after an OOM/crash.

## Mechanism (confirmed against the code)

Per-round checkpoints are saved under the `checkpoints/` subdirectory as `vae_encoder_round_XX.keras` etc. But `TrainingPipeline.load_models(tag, dir=None)` (`src/aetherscan/train.py:5784-5854`) searches the **models root** when `--load-dir` is omitted:

```python
# train.py:5793-5821 (abridged)
if dir is not None:
    base_dir = os.path.join(self.config.model_path, dir)   # e.g. checkpoints/
else:
    base_dir = self.config.model_path                       # <-- ROOT, not checkpoints/
encoder_path = os.path.join(base_dir, f"vae_encoder_{tag}.keras")
...
if not (os.path.exists(encoder_path) and os.path.exists(decoder_path)):
    logger.warning(f"No models tagged as '{original_tag}' in {base_dir}")
    logger.warning(f"Looking for latest tag in {base_dir} instead")
    tag = get_latest_tag(base_dir)      # <-- SILENTLY loads a different, unrelated tag
    logger.info(f"Tag '{original_tag}' not found. Loading latest model with tag: '{tag}'")
    ...
logger.info(f"Successfully loaded models from {base_dir} with tag '{tag}'")   # reports success
```

So `--load-tag round_01` (no `--load-dir`) → `round_01` not found in the root → **fallback to `get_latest_tag(root)`** → loads whatever stale final model is newest in the root (in the blpc3 run: an unrelated `test_v27`), logging warnings but ultimately "Successfully loaded models".

[...]

PR #272, opened 2026-07-24
feat(cli): simplify tag scheme to {command}_{datetime} + robust in-place resume

### LINKED ISSUE #271: Simplify tag scheme to {command}_{datetime} + robust in-place resume (clean break)
Simplify the `--save-tag` / `--load-tag` scheme and make resume/tag-resolution robust. Confirmed design (maintainer decisions, 2026-07-24). Own PR, **before** the v1.0.0 release run (so the release trains under the new scheme). Bundles the deferred #270 items 1–2 (training-plots-dir helper + its unit test) since they're the same `save_tag`-in-`train.py` surface.

## New tag scheme: `{command}_{DATETIME}`

`command ∈ {test, train, inf, bench}`, `DATETIME = YYYYMMDD_HHMMSS` auto-filled **at run time** (fixes the current import-time-frozen `datetime.now()` default at `config.py:440`).

- **`--save-tag`** accepts only `test | train | inf | bench` (the prefix). Runtime → `{prefix}_{DATETIME}`.
  - Omitted → **default to the subcommand**: `train` → `train_{DATETIME}`, `inference` → `inf_{DATETIME}`.
  - `round_XX` is **rejected** on `--save-tag`.
- **`--load-tag`** takes the **full** tag `{cmd}_{DATETIME}` **or** `round_XX` (intermediate checkpoints kept as-is).
- **Clean break:** old formats (`test_vX`, `final_vX`, bare `YYYYMMDD_HHMMSS`) are **rejected** by the new sanitizer on both flags. Pre-refactor on-disk artifacts become load-incompatible via the CLI (acceptable — release trains fresh under the new scheme).

## Resume = in-place via `--load-tag`

Re-running `--save-tag train` mints a fresh datetime (new run), so resume goes through `--load-tag {full_tag}`:
- The run **adopts `{full_tag}` as its save-tag** and resumes from that tag's `run_state_{tag}.json` (same DB rows / artifacts / completed_rounds).
- **Load-fallback ladder, current-tag-only:** try the tag's final weights (`vae_encoder_{tag}.keras`) → else its **latest `round_XX`** checkpoint under that tag → else **fail loudly**.
- **Remove the cross-run `get_latest_tag` fallback** (`train.py:219`, whose own comment warns it "could resume from a stale, unrelated model"). Never load a different run's tag.

## Release / HF-download resolution (`hf_hub.py`)

`final_vX` is gone, so rework the download-default resolution (used by bare `inference` to fetch release weights):
- Release tags (`vX.Y.Z`) remain the primary/top-ranked key (unchanged).
- Replace the `_FINAL_TAG_PATTERN` training-tag fallback. **Proposed (open for review):** drop the training-tag fallback and require a `vX.Y.Z` release tag for a no-artifact inference download; if a training fallback is kept, rank the new `{cmd}_{DATETIME}` tags by datetime. Update `hf_tag_release.py` (`--save-tag final_v1` example) and `docs/RELEASE.md` accordingly.

## Scope / touchpoints (to be finalized by the code map)

- `cli.py`: `_TAG_PATTERN` / `_ROUND_TAG_PATTERN` → new save/load patterns; `--save-tag`/`--load-tag` help; the three validation sites (~L1992/2040/2372); subcommand default; runtime datetime append.
- `config.py`: `save_tag` default (L440) → runtime resolution; HF-revision-default doc (L429).
- `train.py`: `get_latest_tag` + resume/load-fallback (L219–350), run_state keying, resume-in-place adoption of the load tag. **+ the #270 `_training_plots_dir` helper + collapse the ~15 plot-path sites.**
- `hf_hub.py`: `_FINAL_TAG_PATTERN` + resolution ladder.
- `tag_guards.py`: collision guards under the new pattern.
- `main.py` / `data_generation.py` / `run_state.py`: `checkpoint.save_tag` usages.
- Tests: tag-pattern/validation, resume, hf-resolution, tag_guards.
- Docs: `README` (Usage Examples + regenerated CLI Reference), `docs/CONFIG_AND_CLI.md`, `docs/TRAINING_PIPELINE.md`, `docs/RELEASE.md`, `KNOWN_ISSUES.md` if needed.

## Side benefit

Dissolves the `final_v1` tag-collision blocker (2100 stale `training_stats` rows on bla0) — the release no longer uses `final_v1`.

zachtheyek:
## Summary

Simplifies the `--save-tag` / `--load-tag` scheme to **`{command}_{YYYYMMDD_HHMMSS}`** and makes resume/tag-resolution robust. Design confirmed with the maintainer (see #271). **Release-critical** — it retires `final_vX`, so the release run trains under the new scheme (`train_{datetime}`), which also dissolves the `final_v1` tag-collision blocker.

### New scheme
- **`--save-tag`** takes a bare prefix `test|train|inf|bench`; the datetime is stamped **at run time** by `cli.resolve_save_tag()` (in `main()`, before `init_logger`, so log / config / artifacts share one datetime — this also fixes the old import-time-frozen `datetime.now()` default). Omitted → defaults to the subcommand (`train`→`train_…`, `inference`→`inf_…`). `round_XX` is rejected on `--save-tag`.
- **`--load-tag`** takes a full `{cmd}_{datetime}` tag or `round_XX`. **Clean break** — `test_vX` / `final_vX` / bare-datetime are rejected by the new patterns.

### Resume = in-place via `--load-tag`
Re-running `--save-tag train` mints a fresh datetime, so resume goes through `--load-tag {full_tag}`: the run **adopts that tag** and resumes from its `run_state` / DB rows / artifacts. The load-fallback is **current-tag-only** — try the run's final weights → else its last completed round (per manifest) → else **fail loudly**. The cross-run `get_latest_tag()` fallback (the issue-#142 footgun that "could resume from a stale, unrelated model") is **removed**.

### HF download resolution
`select_default_revision` drops the `final_vX` training-tag fallback: a no-artifact `inference` download now **requires a blessed `vX.Y.Z` release tag** (maintainer-approved). `hf_tag_release.py` / `RELEASE.md` updated — bless with the actual `train_{datetime}` run tag.

## Touchpoints
`cli.py` (patterns + `resolve_save_tag` + validation + help + apply), `config.py` (`save_tag` default → runtime-resolved `None`), `main.py` (resolve once before the logger), `train.py` (remove `get_latest_tag`, rewrite `_resolve_load_tag` + the resume flow), `hf_hub.py` (drop `_FINAL_TAG_PATTERN`). Tests: new pattern + `resolve_save_tag` cases, `get_latest_tag` tests removed, `_resolve_load_tag` default-fallback now asserts fail-loudly. Docs: ARCHITECTURE / INFERENCE / RELEASE behavior + README CLI regen + all usage/runbook/script examples.

## For review (please eyeball)
- **The resume flow** (`train.py`, `run_training_pipeline` load block): full-tag resume-in-place → final weights → else `round_{start_round-1}` per manifest → else fail loudly. Round checkpoints are still flat under `{model_path}/checkpoints/` (single-run at a time via startup archiving) — the manifest (which is save_tag-scoped) drives *which* round, so resume stays run-specific. (Per-run-scoped checkpoint dirs would be a larger follow-up.)
- **`resolve_save_tag`** adoption rule: a **full** `--load-tag` is adopted; a `round_XX` `--load-tag` is not (seeds a fresh run).

## Test plan
- [x] `ruff` + `pre-commit` clean; GPG-signed commits.
- [x] Core tag logic validated by direct import (patterns, `resolve_save_tag`, `save_tag is None` default) — the full pytest suite can't run on the `.nosync` dev box (TF import deadlock), so CI is the gate.
- [ ] CI (`pytest -m "not gpu and not cluster"` on 3.10/3.11/3.12) green.
- [ ] Reviewer: the resume flow + the "release tag required for no-artifact download" change.

Closes #271

### DIFF
--- a/.claude/skills/aetherscan-repo-context/SKILL.md
+++ b/.claude/skills/aetherscan-repo-context/SKILL.md
@@ -35,10 +35,10 @@ CLI flags are identical between the two paths; only the launcher differs. `PYTHO
 
 ```bash
 # Train (container, canonical)
-./utils/run_container.sh python -m aetherscan.main train --save-tag final_v1
+./utils/run_container.sh python -m aetherscan.main train --save-tag train
 
 # Train (conda source, Ampere)
-PYTHONPATH=src python -m aetherscan.main train --save-tag final_v1
+PYTHONPATH=src python -m aetherscan.main train --save-tag train
 
 # Inference from a pre-processed .npy
 ./utils/run_container.sh python -m aetherscan.main inference \
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -10,9 +10,9 @@ Aetherscan: Breakthrough Listen's deep-learning SETI pipeline. Two-stage ML (Bet
 
 ```bash
 # Canonical: NGC container (only option on Blackwell)
-./utils/run_container.sh python -m aetherscan.main {train|inference} --save-tag final_v1
+./utils/run_container.sh python -m aetherscan.main {train|inference} --save-tag train
 # Alternative: conda env (Ampere only) — prefix with PYTHONPATH=src
-PYTHONPATH=src python -m aetherscan.main {train|inference} --save-tag final_v1
+PYTHONPATH=src python -m aetherscan.main {train|inference} --save-tag train
 # Lint + format (also enforced via pre-commit)
 ruff check src/ && ruff format src/
 ```
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ The `AETHERSCAN_*` paths are bind-mounted 1:1 between host and container, so the
 
 ```bash
 ./utils/run_container.sh python -m aetherscan.main {train|inference} \
-  --save-tag final_v1
+  --save-tag train
 ```
 
 The `utils/run_container.sh` wrapper auto-detects whether `apptainer` or `singularity` is on PATH (Apptainer wins when both are present), sets `--nv` for GPU passthrough, and binds the repo + `AETHERSCAN_{DATA,MODEL,OUTPUT}_PATH` 1:1 between host and container so absolute paths persisted in the DB stay valid across both. `PYTHONPATH` is set automatically inside the container — no inline prefix needed.
@@ -201,7 +201,7 @@ Multiprocess worker pools inherit the values via `os.environ` as usual.
 
 ```bash
 PYTHONPATH=src python -m aetherscan.main {train|inference} \
-  --save-tag final_v1
+  --save-tag train
 ```
 
 `PYTHONPATH=src` makes the `aetherscan` package importable from `src/` without a `pip install -e .` step. No inline `KEY=VALUE` prefix is needed for Slack credentials — the `.env` auto-load runs before any worker process is spawned, so `os.environ` inheritance to multiprocess pools is automatic.
@@ -243,7 +243,7 @@ PYTHONPATH=src python -m aetherscan.main train
     --num-training-rounds 20 \
     --epochs-per-round 100 \
     --curriculum-schedule exponential \
-    --save-tag test_v1
+    --save-tag test
 
 # Source
 PYTHONPATH=src python -m aetherscan.main train \
@@ -251,7 +251,7 @@ PYTHONPATH=src python -m aetherscan.main train \
     --num-training-rounds 20 \
     --epochs-per-round 100 \
     --curriculum-schedule exponential \
-    --save-tag test_v1
+    --save-tag test
 ```
 
 **Resume from checkpoint**
@@ -261,13 +261,13 @@ PYTHONPATH=src python -m aetherscan.main train \
 ./utils/run_container.sh python -m aetherscan.main train \
     --load-dir checkpoints \
     --load-tag round_10 \
-    --save-tag test_v1
+    --save-tag test
 
 # Source
 PYTHONPATH=src python -m aetherscan.main train \
     --load-dir checkpoints \
     --load-tag round_10 \
-    --save-tag test_v1
+    --save-tag test
 ```
 
 > [!WARNING]
@@ -279,12 +279,12 @@ PYTHONPATH=src python -m aetherscan.main train \
 # Container
 ./utils/run_container.sh python -m aetherscan.main train \
     --gpu-memory-limit-mb 14000 \
-    --save-tag test_v1
+    --save-tag test
 
 # Source
 PYTHONPATH=src python -m aetherscan.main train \
     --gpu-memory-limit-mb 14000 \
-    --save-tag test_v1
+    --save-tag test
 ```
 
 **Watching the live dashboard from your local browser (SSH port forwarding)**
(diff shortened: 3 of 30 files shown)


zachtheyek:
Thorough pass — and the `train.py:994` finding was a **real bug** that would have made every resume-in-place fail. All addressed in `eaa0303`:

1. **`_start_round`-clobbering (critical)** — fixed via your option (2). `_init_run_state` now computes `resume_in_place = load_tag is not None and load_tag == save_tag` and **excludes it from `explicit_checkpoint`**, so resume-in-place falls through to the manifest-driven branch (`_start_round = state.resume_round`) and `completed_rounds` is not wiped. So `train --save-tag train --load-tag train_20260101_120000` with rounds 1–N done + no final weights now correctly loads `round_N`. You're right that resume-in-place is a *continuation*, not an override.

2. **Missing test** — added `TestResumeInPlace::test_manifest_rounds_drive_start_round`: writes a manifest (`completed_rounds=[1,2,3]`, fingerprint-matched) for a resume-in-place tag with no final weights, calls `_init_run_state`, and asserts `_start_round == 4` and the rounds survive. It fails on the pre-fix code (would assert `1` / `[]`), so it pins the regression.

3. **Dead round check (`train.py:981`)** — dropped. `resolve_save_tag` never adopts a `round_XX` load-tag, so `load_tag == save_tag` already implies a full tag; the guard is now just `load_tag is not None and load_tag == save_tag`.

4. **Unknown-command → unloadable tag (`cli.py:45`)** — `resolve_save_tag` now raises `ValueError` for a command outside `{train, inference}` (when `--save-tag` is also omitted) rather than emitting a `run_…` tag that `_LOAD_TAG_PATTERN` would reject.

5. **Silent `--save-tag` drop (`cli.py:60`)** — `main()` now logs an INFO line right after `init_logger` when a full `--load-tag` adopts its tag over an explicitly-provided `--save-tag` prefix, so the ignored save-tag isn't a mystery.

The separate 2 CI failures (the `select_default_revision` tests asserting the removed `final_vX` fallback) were fixed in `a2d878f`. CI re-running on `eaa0303`.

@claude

zachtheyek:
Pushed 4 commits (through `bc2aee4`); CI is green on 3.10 / 3.11 / 3.12. The review landed on the pre-rebase HEAD, so several notes were already resolved by the rebase + resume-fix (`d55fd9f`); here's the point-by-point, plus a fresh-eyes pass that turned up a real regression and some doc drift.

### Original review notes — all addressed

1. **`train.py` — resume-in-place `_start_round` clobber (the manifest fallback can't fire).** Fixed in `d55fd9f`: `_init_run_state` now treats resume-in-place (`load_tag == save_tag`) as a *continuation* (`explicit_checkpoint = False`), so the manifest's `resume_round` drives `_start_round` and `completed_rounds` is preserved rather than wiped. I additionally extracted the load decision into `_resume_load_plan()` and added `TestResumeLoadPlan` (`d14b028`), so the round-checkpoint fallback this note was about is now directly unit-tested — see note 5.

2. **`train.py:981` — dead `round_\d+` defensive check.** Gone. The resume-in-place test is now just `load_tag is not None and load_tag == save_tag`; a full `--load-tag` can never be `round_XX` because `resolve_save_tag` never adopts a round tag, so the check was redundant and was dropped.

3. **`cli.py:69` — silent `--save-tag` drop when a full `--load-tag` is also passed.** `main()` now logs it explicitly ("Resuming in place under `--load-tag` … the provided `--save-tag` … was ignored").

4. **`cli.py:45` — `.get(command, "run")` emits an unloadable tag.** `resolve_save_tag` now raises `ValueError` for an underivable prefix instead of silently producing `run_…`. Fresh-eyes follow-up: that raise was reachable *unguarded* from `main()` on a bare `python -m aetherscan.main` (subparser wasn't `required`), so a no-subcommand invocation crashed with a traceback before the logger existed instead of printing help. Fixed by marking the subparser `required=True` (`407c813`), which restores the graceful help/exit via argparse's usage error. Added a `resolve_save_tag` `ValueError` test.

5. **`test_train_utils.py:88` — resume-in-place load fallback untested.** Covered from both ends now: `TestResumeInPlace` exercises `_init_run_state` (manifest drives `_start_round`, rounds not wiped), and `TestResumeLoadPlan` exercises the load block via the extracted `_resume_load_plan()` — asserting fall-back-to-`round_03`, fail-loud when there's no final and no rounds, load-final-when-present, and no-op on a fresh run.

### Additional fixes from the fresh-eyes pass (`bc2aee4`)

- **CI (was red):** `config.checkpoint.save_tag` now defaults to `None` (resolved at runtime), but `test_bandpass_debug_plot_saved` built the config directly and hit `os.path.join(..., None, ...)`. Resolved a realistic `save_tag` in the test's `_configure` to mirror production (`d802333`).
- **Public HF model card:** a training-tag card no longer claims its weights download "by default" — a no-artifact inference download resolves to the latest `vX.Y.Z` release tag, never a training tag.
- **Resume-semantics doc drift** (`TRAINING_PIPELINE.md`, `INFERENCE_PIPELINE.md`, a `main.py` retry comment): a full process relaunch now mints a fresh datetime tag; resume is explicit via `--load-tag {full-tag}` (train) or `--preprocess-output-dir` (inference), not "re-run the identical command".
- `load_models` docstring no longer references the removed `get_latest_tag` "latest tag present" fallback; agent docs (`CLAUDE.md`, `SKILL.md`) and README examples moved off `final_v1` / the invalid `run_v1`.

Requesting a confirmation pass on the resume path and the doc changes. @claude

zachtheyek:
Both fixed in `e5db361`, plus a third analogous drift a full-tree sweep turned up:

- **`aetherscan.def` `%help`** — `--save-tag test_v1` → `test`.
- **`logger.py`** — `Logger.__init__` / `init_logger` docstrings + the inline comment no longer describe a "config default timestamp tag" or `init_logger` running before `--save-tag` is applied; `main()` resolves the `{command}_{datetime}` tag onto the config before calling `init_logger` (code was already correct).
- **`tag_guards.py`** module docstring — the "default (datetime) tags immune, guards only fire for explicit tags" framing is moot now every resolved tag carries a fresh datetime; reworded to match (same class as the `INFERENCE_PIPELINE.md` fix).

A full sweep for `--save-tag`/`--load-tag` old-scheme values, `get_latest_tag`, `_TAG_PATTERN`, and "default timestamp" wording across all tracked files is now clean. CI green. Requesting a final confirmation. @claude

PR #283, opened 2026-07-26
Live-run hardening: input-pipeline, DB writer, GIF tail, unified seeding, monitor/logger fixes, latent-representation overhaul (#276–#282)

zachtheyek:
## Summary

Implements all seven open engineering issues from the 2026-07-26/27 live-run findings (#276–#282) on one branch, in issue order, with one focused commit per issue plus a benchmark-numbers commit and a docs-sync commit. Every issue's maintainer constraints are honored; both perf issues were **audited on blpc3 first** (harnesses + numbers committed under `benchmarks/`), and blpc3 was used as the sole test bed — **the bla0 release run was never touched**.

### #276 — GPU starvation (audit first, then fix)
`benchmarks/bench_input_pipeline.py` measures the REAL memmap → tf.data → distribute → train-step path (the gap `bench_gpu.py` documents), with a verbatim pre-#276 legacy builder for before/after, a `--gil-load` contention knob, `--deterministic-ops` isolation, and a `--profile` TF-profiler hook. Full numbers and conclusions in `benchmarks/README.md`.

**TF profiler (the evidence the issue mandated):** XPlane compute-stream occupancy is **6.3% legacy / 8.3% current — the GPUs are idle >90% of the wall clock**, while doing *exactly* the kernel work `bench_gpu` predicts (3,253 cad/s per GPU measured from the trace vs 2,986 synthetic). Per-GPU kernel time is identical between variants (4.72 s); only the waiting shrinks (75.7 s → 57.2 s window for the same work).

**GIL contention is the dominant mechanism, measured end-to-end:** one background drainer-shaped Python thread costs **54%** of legacy 5-GPU training throughput; two cost **91%** (an 11× collapse). Caveat stated in the doc: the hog is a synthetic proxy, uncalibrated against the real drainer's burst rate — it establishes mechanism and sensitivity, not the exact live magnitude.

**The memmap copy is the NEXT wall, not the current one:** at the throughput actually observed it runs **8.4× under** its own 6.97 GB/s capability — the GIL is what binds today. But a *fully fed* 5-GPU consumer would draw **9.59 GB/s** (trace-derived ceiling), and one thread supplies **0.73×** of that, so the copy becomes co-limiting once the GIL ceiling is removed. Both halves matter: don't optimize the copy now, don't assume it scales when you do.

[...]

Issue #279, opened 2026-07-26
[BUG]: unify random state under one config field — inference is unseeded, two disjoint seed roots, 6 hard-coded literals


zachtheyek:
## Summary

The pipeline has **two disjoint seed roots with opposite defaults**, **six hard-coded `11` literals**, and an **inference path that is not seeded at all**. The maintainer wants a **single unified field in `config.py`** that everything else derives from.

The codebase already agrees: there are **8 in-tree `# NOTE: use a global config seed instead of hard-coding` markers** ([train.py:2717](../blob/master/src/aetherscan/train.py#L2717), [:4482](../blob/master/src/aetherscan/train.py#L4482), [:4715](../blob/master/src/aetherscan/train.py#L4715), [:4761](../blob/master/src/aetherscan/train.py#L4761), [:5778](../blob/master/src/aetherscan/train.py#L5778), [:5784](../blob/master/src/aetherscan/train.py#L5784), [:6020](../blob/master/src/aetherscan/train.py#L6020), [:6152](../blob/master/src/aetherscan/train.py#L6152)). This issue is the consolidated plan to discharge them.

Findings below are from a read-only audit, adversarially verified; the headline item was re-verified by hand.

---

## 🔴 Finding 1 (HIGH) — inference is genuinely non-reproducible

This is a **science-output** problem, not a convenience one, and it is the reason to prioritise this issue.

- [`inference.py:427`](../blob/master/src/aetherscan/inference.py#L427): `_, _, z = self.encoder(reshaped, training=False)` — inference consumes the **stochastic sampled `z`**, not `z_mean`.
- [`models/vae.py:51`](../blob/master/src/aetherscan/models/vae.py#L51): `epsilon = tf.random.normal(shape=(batch, dim))` — an **unseeded** stateful op. `training=False` does not disable it; `Sampling` always samples.
- `grep -nE "set_seed|tf\.random|seed" src/aetherscan/inference.py` → **zero hits**. `tf.random.set_seed` is called *only* in `TrainingPipeline.__init__` ([train.py:896-898](../blob/master/src/aetherscan/train.py#L896)); `InferencePipeline` never seeds anything. `tf_deterministic_ops` is likewise training-only.
- The inference subparser exposes **0** seed/determinism flags (the train parser has 2), so a run cannot be pinned from the CLI either.
[...]

Issue #278, opened 2026-07-26
[PERF]: latent-space GIF stage is a ~24-29 h tail — benchmark first, then parallelize frame rendering (not the UMAP fit)


zachtheyek:
## Summary

The latent-space GIF stage (`TrainingPipeline.plot_latent_space_gif`, [train.py:4309](../blob/master/src/aetherscan/train.py#L4309)) takes an estimated **~24–29 hours**. It runs **once, in the final `vae_plots` stage**, so the release training run pays it in full at the very end — **larger than the ~1.4 h MP-SHAP tail** that was previously treated as the dominant end-of-run cost.

Measured on blpc3's `test_20260724_190816` run (read-only, from the run log).

**Constraints for this work (maintainer):**
- **Benchmark first — confirm the exact bottleneck before implementing anything.** Same discipline as the SHAP work (#258/#259) and #276. The decomposition below is inferred, not instrumented.
- **Do NOT drop `random_state` where it is currently set.** Determinism stays. UMAP's `n_jobs=1` single-threading is an accepted consequence, not a target.

## Where the time goes

The sweep is `latent_viz_umap_n_neighbors=[5,15,30,50]` × `latent_viz_umap_min_dist=[0.0,0.1,0.5]` = **12 combos**, each producing an **obs-level and a cadence-level** GIF ⇒ **24 UMAP fits + 24 GIFs**, 500 frames each.

Per-combo timings from the log:

| stage | obs-level | cadence-level |
| --- | --- | --- |
| UMAP fit | ~50–70 s | ~50–70 s |
| transform + render + assemble | **~61 min** | **~56 min** |

Observed throughput: 6 of 24 GIFs in ~7.2 h ⇒ ~72 min/GIF ⇒ **~24–29 h total**.

**The UMAP fit is ~1.5 % of the cost. It is not the bottleneck.**

### The bottleneck is very likely matplotlib, not UMAP

Obs-level transforms **23,040 vectors/frame** (11.52 M pooled); cadence-level transforms **3,840/frame** (1.92 M) — **6× fewer** — yet cadence is only ~8 % faster overall. Solving the two equations:

```
T_obs     + R = 61 min
T_obs / 6 + R = 56 min   ⇒   T_obs ≈ 6 min,  R ≈ 55 min
```

⇒ **~55 min of rendering vs ~6 min of transform (obs) / ~1 min (cadence)** — about **6.6 s per frame** of matplotlib.

[...]

PR #288, opened 2026-07-28
perf: training optimization round 2 — 21.9× generation, parallel GIF sweep, query-window fix, A/B-gated fp16/bf16 flags

### LINKED ISSUE #287: [PERFORMANCE]: Training pipeline optimization round 2 — the post-#283 walls (generation, plot queries, GIF sweep, GPU thread mode, A/B-gated fp16/bf16)
## Context

PR #283 made the training hot path Python-free (step throughput 8.5×, quiet epochs 48.5→5.8 s at A/B scale), which moved the wall: a production-scale 2-round run on blpc3 (2026-07-27, `train_20260727_204719`, killed by the maintainer to pivot to this work) measured **round data generation at ~6.2 h/round (37.4 min × 10 chunks) with the GPUs at 0%** — all 32 pool workers pegged at ~94.5% CPU, ~97% user-mode. With post-#283 training at ~1–2 h/round, generation dominates the wall clock on blpc3-class hosts, and the end-of-run tails (UMAP GIF sweep ~1.7–1.9 h serial) plus per-plot DB query scans grow with campaign length.

A 12-agent optimization pass over the training pipeline (adversarially verified through science-purity / already-tried / Amdahl / stack-risk lenses, against the pinned null results in `benchmarks/README.md`) produced a ranked plan. This issue tracks the implementation.

## Scope (all byte-identical or default-off)

1. **Generation wall**: remove the per-injection `gc.collect()` in `new_cadence` (~2.5M full collections/round in TF-laden workers; measured 23 ms/call vs 4.5 ms for the rest of the function); draw-first `create_true_double` (the intersection test is a pure function of pre-array RNG draws — at p≈0.42 acceptance, ~41% of setigen injections were computed then discarded); one msync per array instead of ~23,400 per-task full-mapping msyncs (same `.done`-manifest durability contract). Gate: a seeded A/B harness (`benchmarks/bench_datagen.py`) driving the real pooled path with per-array sha256s — byte-identical or it doesn't land.
2. **Injection-plot query windows**: `plot_injection_stats` issues ~165 round-scoped queries per call, but `idx_injection_stats_filter` leads with `(tag, timestamp)` and lacks `round_number` — the run-wide window re-scans the tag's whole history per query (measured 10.5× slower at 12M rows, quadratic over a campaign). Tighten via a whole-partition MIN/MAX span aggregate (superset bound ⇒ provably identical result sets).
3. **UMAP GIF sweep**: parallelize the 24 independent (n_neighbors × min_dist × level) combos across forkserver workers — each combo consumes only its own derived random_state (#279), so process isolation preserves every output byte (disjoint from the #278-rejected within-fit ideas, which stay rejected). ~1.7–1.9 h → ~10–15 min expected.
4. **`TF_GPU_THREAD_MODE=gpu_private`** behind config, aimed at the measured ~7.6% input h2d/scheduling interference (the one step-level residual without a pinned null); default validated by a step-ladder A/B before merge.
5. **Two A/B-gated numerics levers, DEFAULT OFF** (maintainer-approved in scope, gated on val-metric parity): `training.round_array_dtype="float16"` (halves the ~294.5 GB round footprint, gather volume, and page-cache working set; manifest-gated on every reuse/resume path) and `beta_vae.mixed_precision` (keras `mixed_bfloat16` with fp32 islands for the z-heads/Sampling/decoder-sigmoid so all loss math stays fp32; no loss scaling; no new collectives). Flag flips are a follow-up gated on: 3 seeds × 2 arms, val AUC within max(2σ, 0.002), losses/recalls within 2σ, same active-dim count, zero NaN trips.

## Deferred with rationale (recorded so they are not silently resurrected)

RF-dataset pre-generation on the producer, direct-numpy injection (bypassing setigen Frames), SHAP-stage overlap, pool thread-pinning, fused moments — the measured 21.5× generation result collapsed their absolute value to ~minutes each.

## Acceptance criteria

- `bench_datagen` A/B: byte-identical checksums vs master + the measured speedup ladder published in `benchmarks/README.md`.
- Full unit suite green in the NGC container on blpc3; new unit pins for the RNG-replay equivalence, manifest dtype gates, sweep byte-identity, and bf16 dtype islands.
- Both numerics flags verifiably inert at their defaults (regression pins).
- Docs synced (TRAINING_PIPELINE, CONFIG_AND_CLI, DATABASE, BENCHMARKING, benchmarks/README).

zachtheyek:
Closes #287

## Summary

Training-pipeline optimization round 2 — the walls left standing after #283 moved the hot path off Python. Everything here is either **byte-identical by construction and by checksum**, or lands **behind a default-off flag** whose default reproduces the pre-flag pipeline byte-for-byte.

**Headline measurement** (`bench_datagen.py` A/B ladder; blpc3, 8192 samples × 3 arrays, 32 workers, `--preload-tf`, seed 11):

| arm | wall | speedup | sha256s (all 7 outputs) |
|---|---|---|---|
| A: master (`fa08b0f`) | 282.0 s | 1.0× | baseline |
| B: + gc removal | 18.6 s | 15.2× | identical to A |
| C: + draw-first (as shipped) | 13.1 s | **21.5×** | identical to A |

Projected production round generation on blpc3: **~6.2 h → ~20 min** (projection from the benchmark ratio; the next full-scale run's `data_generation` stage timers are the check). The live-run evidence behind the ranking: 32 producer workers at ~94.5% CPU / ~97% user-mode while the GPUs sat at 0% for ~6.2 h/round.

## What changed

- **`gc.collect()` per injection removed** (`data_generation.new_cadence`): a full generational collection ran ~2.5M times/round inside pool workers carrying the whole TF import graph — measured ~23 ms/call vs ~4.5 ms for the entire rest of the function. Refcounting already frees the setigen Frame; the per-chunk collect stays.
- **Draw-first `create_true_double`**: the intersection-acceptance test only needs `slope_pixel`/`y_intercept` — pure functions of the pre-array RNG draws — but the retry loop materialized BOTH setigen injections before testing (~41% of all injections discarded at p≈0.42). `new_cadence` is split into `_draw_signal_params` (consumes the exact 4-draw sequence) + `_inject_drawn_signal` (consumes no RNG — pinned by test); the loop replays identical per-attempt draws and materializes only the accepted pair. Byte-equivalence pinned by a verbatim pre-split reference loop over 4 seeds, including final RNG stream states.
- **One msync per array** instead of ~23,400 per-task full-mapping msyncs per round; the `.done` manifest is still written strictly after every array is flushed (same durability contract).
- **`bench_datagen.py`**: the seeded A/B harness gating all of the above (real pooled path, per-array sha256s, `--preload-tf` to mirror the producer workers' import graph).
- **Injection-plot query windows** (`plot_injection_stats` + new `Database.query_injection_stat_time_span`): ~165 round-scoped queries per call re-scanned the tag's whole `(tag, timestamp)` index partition because `round_number` is not in the index (measured 10.5× at 12M rows, quadratic over a campaign). One whole-partition MIN/MAX aggregate — a deliberate superset bound (superseded + non-finite rows included) — tightens the window; intersection can only narrow scans, never change a result set. Unit tests pin the superset-bound property.
- **UMAP GIF sweep parallelized across whole combos** (`latent_gif.run_umap_gif_sweep`): the 24 (n_neighbors × min_dist × obs/cadence) combos ran strictly serially at ~95% single-core (~1.7–1.9 h/run) even after #278 parallelized frame rendering. Each combo is an independent fit consuming only its own derived random_state (#279 sub-keys), reads shared inputs read-only, writes distinct files — forkserver workers (empty preload, pinned native threads incl. numba, the `shap_parallel` pattern) preserve every output byte. **Disjoint from the #278-rejected within-fit ideas**: batched transforms / precomputed-knn reuse stay rejected; per-snapshot transforms stay serial within each combo. Expected ~10–15 min on 24+ cores (not yet measured at production scale). Byte-identity pinned by a slow test comparing serial vs pooled GIF bytes (first real UMAP fit in CI — adds ~1 min of numba JIT once).
- **`gpu.gpu_thread_mode` / `gpu.gpu_thread_count`** → `TF_GPU_THREAD_MODE`/`TF_GPU_THREAD_COUNT` set before GPU-runtime init, aimed at the measured ~7.6% h2d/scheduling interference. Step-ladder A/B result below.
- **Two A/B-gated numerics levers, DEFAULT OFF** (neither makes so much as a policy call when off):
  - `training.round_array_dtype="float16"`: halves the round footprint (~294.5 → ~147 GB), gather volume, and page-cache working set — the lever that matters once the 2-round overlapped working set (~589 GB at full scale) exceeds the hosts' 503 GB RAM. Quantization ≤ 2⁻¹² on [0, 1] log-normed inputs; the gather map's existing `tf.cast` becomes the host-side upcast so the training graph/loss math are unchanged either way. `.done` manifests record dtypes and every reuse/resume path gates on them (legacy manifests read as float32); the cli disk-budget guard is dtype-aware.
  - `beta_vae.mixed_precision`: keras `mixed_bfloat16` set before the model build, with fp32 islands (z_mean/z_log_var heads, `Sampling`, decoder sigmoid) so everything reaching `compute_total_loss` stays fp32; variables/Adam/step accumulators fp32; no loss scaling; no new collectives (the Blackwell NCCL constraint is untouched). `bench_gpu.py --mixed-precision` added for the Phase-0 gate.
  - **The gate for flipping either flag**: 3 seeds × 2 arms at scaled shape; val AUC within max(2σ, 0.002), losses/recalls within 2σ, identical active-dim count, zero NaN-guard trips. Phase-0/1 numbers from this branch are below; the bf16 **A4000 VRAM leg is deliberately deferred** — bla0 is running the Phase-2 release model and must not be touched.

## Validation (blpc3, NGC container)

- **Full unit suite** (NGC container): **937 passed** (master baseline on this host: 909), 5 deselected (gpu/cluster), 5m52s — includes the new pins: RNG-replay equivalence over 4 seeds incl. final stream states, no-RNG materialization contract, manifest dtype gates (incl. legacy manifests), GIF-sweep byte-identity (serial vs pooled GIF bytes), bf16 dtype islands + an all-fp32 regression pin at the default policy.
- **Generation ladder**: master 282.0 s → branch tip **12.9 s (21.9×)**, sha256s identical to master across all 7 outputs — the tip number includes the msync change (bytes unchanged, only when they flush).
- **fp16 mechanics smoke** (container, CPU): a float16 round generates, its `.done` manifest dtype-gates reuse, `_as_cpu_tensor` dlpack-wraps it zero-copy as float16, and gather+cast yields float32.
- **bf16 Phase-0 gate** (`bench_gpu.py`, 1× RTX PRO 6000, batch 128, K=12): 2983 → 3443 agg cadences/s (**1.15×**, above the 1.1× abort threshold) with bench-loop peak VRAM 23.42 → 12.59 GB. The production tf.range loop peaks far lower than the bench's unrolled loop; the **A4000 VRAM leg is deferred until bla0 frees** (Phase-2 release run in flight — not to be touched).
- **TF_GPU_THREAD_MODE step A/B** (`bench_input_pipeline` step mode, 5 GPUs, 3 reps/arm): default mean 7159 (6933–7435) vs `gpu_private` mean 7400 (7271–7622) = **+3.4% mean, no regression**. The flipped default stays with these numbers as the evidence; `gpu.gpu_thread_mode="global"` restores the old behavior.
- **The val-metric A/Bs that gate FLIPPING the two numerics flags run next on blpc3** (3 seeds × 2 arms each at scaled shape, per the gate above) — they can run concurrently with review; the flags stay off in this PR regardless of their outcome.

## Deferred with rationale

RF-dataset pre-generation on the producer, direct-numpy injection bundle, SHAP-stage overlap, pool thread-pinning, fused moments — the 21.5× generation result collapsed their absolute value to ~minutes each. Recorded in `benchmarks/README.md` so they are not silently resurrected.

## Out of scope, flagged for the maintainer

The 12-agent pass surfaced (M5, unverified): the documented L1/L2 regularizers on the VAE appear to be **silently inactive** (kernel-count check pending). That is a potential correctness/documentation issue, not a performance one — deliberately not touched here; happy to file it separately once the kernel-count check runs.

## Docs

`benchmarks/README.md` (bench_datagen + producer follow-up section with the full evidence chain), `docs/TRAINING_PIPELINE.md` (second-pass performance-engineering continuation + the A/B-gated flags), `docs/CONFIG_AND_CLI.md`, `docs/DATABASE.md`, `docs/BENCHMARKING.md`. No `cli.py` argparse surface changed → README CLI Reference unchanged (verified imports stay stdlib-only).

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -20,6 +20,9 @@ python benchmarks/bench_rf.py                    # Random Forest stage: latent p
 # Container-only (need the full TF/umap stack; bench_input_pipeline's step mode needs GPUs):
 ./utils/run_container.sh python benchmarks/bench_input_pipeline.py --mode iterate --variant current
 ./utils/run_container.sh python benchmarks/bench_latent_gif.py --mode all
+# Generation-path A/B (real pooled producer path; --preload-tf needs TF, so run containerized):
+./utils/run_container.sh python benchmarks/bench_datagen.py --preload-tf \
+    --data-dir /datax/scratch/$USER/bench_datagen
 ```
 
 `bench_gpu.py` is a different animal — it profiles the Beta-VAE on a real GPU (throughput +
@@ -43,6 +46,7 @@ comparing against the baselines.
 | `bench_rf.py` | `RandomForestClassifier` fit + `predict_proba` and the `prepare_latent_features` reshape (sklearn, CPU) | Second-stage RF training + inference (`train.train_random_forest`) |
 | `bench_input_pipeline.py` | The REAL memmap → tf.data → distribute → train-step input path (gather / iterate / step modes; legacy vs current builder; `--gil-load` contention knob) — **container-only; step mode needs GPUs** | The training input pipeline `bench_gpu.py` deliberately excludes (`train.prepare_distributed_train_dataset`) — see [#276 audit](#input-pipeline-audit-276) |
 | `bench_latent_gif.py` | Latent-GIF stage decomposition (UMAP fit / transform / frame render / GIF assembly) with output-equality checks on every candidate optimization — **container-only** | The `vae_plots` latent-GIF tail (`train.plot_latent_space_gif`) — see [#278 audit](#latent-gif-audit-278) |
+| `bench_datagen.py` | Seeded round generation through the REAL pooled `generate_round_to_memmap` path (shared-memory plates, `_init_worker`, per-task seed derivation, batched memmap tasks) with per-array sha256 checksums — the byte-compatibility gate for generation-path changes; `--preload-tf` mirrors the producer workers' TF import graph | The producer wall (`train.round_XX.data_generation`) that `bench_injection.py`'s single-process kernel can't reach — see [Producer/data-generation follow-up](#producerdata-generation-follow-up) |
 | `parse_xplane_occupancy.py` | Post-processor, not a benchmark: per-GPU busy time / occupancy from a `--profile` run's XPlane trace (merged event intervals, strict kernel-time measure) | The "GPUs idle >90%" profiler evidence in the [#276 audit](#input-pipeline-audit-276) and its follow-up |
 
 ## Baseline numbers
@@ -454,6 +458,69 @@ stats.
   backend and float summation order changed); determinism itself is preserved — same seed, same
   split, same per-epoch batch order, byte-identical reruns — and is pinned by unit tests
   (`test_train_datasets.py`, `test_train_accumulation.py`, `test_train_distribution.py`).
+
+## Producer/data-generation follow-up
+
+With the training hot path Python-free, the wall moved. The real-run A/B above already showed
+it at test scale (round-1 training finished long before round-2 generation), and a killed
+production-scale run on blpc3 confirmed it live: ~6.2 h of data generation per round with the
+GPUs at 0% and all 32 pool workers at ~94.5% CPU, ~97% user-mode — compute inside the workers,
+not IO or scheduling. `bench_datagen.py` is the harness this follow-up was gated on: it drives
+the real pooled `generate_round_to_memmap` path (shared-memory plates, `_init_worker`, per-task
+seed derivation, batched memmap tasks) at reduced scale and sha256s every output array, so a
+generation-path change lands only if its checksums match master's byte-for-byte. `--preload-tf`
+imports TensorFlow (CUDA-blanked) into the parent before the pool forks, mirroring the import
+graph the production producer's workers inherit — without it the gc/interpreter costs read
+optimistically low against exactly the regime the live run measured.
+
+Three mechanisms shipped, all byte-identical by construction and by checksum
+(`data_generation.py`):
+
+1. **The per-injection `gc.collect()` in `new_cadence` is gone.** A full generational
+   collection ran once per injection — ~2.5M times per production round — inside pool workers
+   carrying the full TF import graph, and measured ~23 ms per call against ~4.5 ms for the
+   entire rest of the function: the dominant term of the generation wall. Refcounting already
+   frees the setigen Frame's arrays immediately; the per-chunk collect in
+   `generate_round_to_memmap` stays for cycle cleanup.
+2. **Draw-first `create_true_double`.** The intersection-acceptance test is a pure function of
+   the pre-array RNG draws, but the retry loop materialized BOTH full setigen injections before
+   testing — at the measured p≈0.42 acceptance, ~41% of all injections in a production round
+   were computed and discarded. `new_cadence` is split into `_draw_signal_params` (consumes the
+   exact 4-draw RNG sequence) and `_inject_drawn_signal` (consumes no RNG); the retry loop now
+   replays the identical per-attempt draws and materializes only the accepted pair — same
+   order, same values, pinned by byte-compat tests in `test_data_generation.py`.
+3. **One msync per array, not per task.** `_run_memmap_task`'s finally flushed its whole
+   mapping once per task — ~23,400 concurrent full-mapping msyncs per production round, a
+   driver of the #117/#118 chunk-tail stragglers. Replaced by one flush per output array
+   immediately before the `.done` manifest. The durability contract is unchanged: no array
+   bytes are trusted until the manifest exists, and the manifest is written strictly after
+   every array is msync'd.
+
+### A/B ladder (blpc3, 8192 samples × 3 arrays, 32 workers, `--preload-tf`, seed 11)
+
+| arm | wall | speedup | sha256s (all 7 outputs) |
+|---|---|---|---|
+| A: master | 282.0 s | 1.0× | baseline |
+| B: + gc removal | 18.6 s | 15.2× | identical to A |
+| C: + draw-first (as shipped) | 13.1 s | **21.5×** | identical to A |
+
+The msync change lands with arm C — it alters no bytes, only when they are flushed, and is
+covered by the same checksum gate. Projected to production scale on the same 32-core host:
+~6.2 h/round → **~20 min/round**. That is a projection from the benchmark ratio, not a measured
+production round — the next full-scale run's `data_generation` stage timers are the check.
+
+**Deferred with rationale (do not silently resurrect):** RF-dataset pre-generation on the
+producer, a direct-numpy injection bundle (bypassing setigen Frames), SHAP-stage overlap, pool
+thread-pinning, and fused moment computation — the 21.5× result collapsed their absolute value
+to ~minutes each.
+
+**Related fix from the same audit, outside the generation path:** `plot_injection_stats` issues
+~165 round-scoped queries per call against an index that leads with `(tag, timestamp)` and does
+not contain `round_number`, so its run-wide time window re-scanned the tag's whole row history
+on every query — measured **10.5× slower at 12M rows**, and quadratic over a campaign as
+history accumulates. `Database.query_injection_stat_time_span` (one whole-partition MIN/MAX
+aggregate; a superset bound, so intersecting it with a query window can never change a result
+set) now tightens the window to the plotted rounds' actual span — see `docs/DATABASE.md`.
 
 ## Latent-GIF audit (#278)
 
--- a/benchmarks/bench_datagen.py
+++ b/benchmarks/bench_datagen.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+"""
+Round-generation benchmark: the real pooled generate_round_to_memmap path (the producer wall).
+
+bench_injection.py times the single-process new_cadence kernel; this script measures the
+whole round-generation machinery at reduced scale exactly as the producer runs it — a
+worker pool initialized via _init_worker against shared-memory background plates, per-task
+seed derivation from the root seed, batched memmap-writing tasks, per-chunk barriers — and
+then sha256s every output array. Two invocations with identical arguments on different
+code (e.g. master vs a candidate branch) must produce identical checksums: that is the
+byte-compatibility gate for any generation-path optimization, alongside the wall-time delta.
+
+    # arm A (checkout master), then arm B (checkout branch), same args:
+    ./utils/run_container.sh python benchmarks/bench_datagen.py \
+        --n-samples 8192 --workers 32 --seed 11 \
+        --data-dir /datax/scratch/$USER/bench_datagen --output /tmp/arm_a.json
+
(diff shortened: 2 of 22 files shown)


zachtheyek:
Thanks for the pass — all three inline notes worked, point by point (fixes in `5d63f54`):

**1. `_SWEEP_BUNDLE` serial-arm cache leak — fixed.** `run_umap_gif_sweep`'s in-process branch now clears the module-global bundle cache in a `finally`, so the caller's pre-sweep `del` + `gc.collect()` isn't undone by a lingering multi-hundred-MB cache. The pooled arm is deliberately unchanged — its cache lives in short-lived forkserver workers, which is the reuse the cache exists for.

**2. `_draw_signal_params` annotation (`dict[str, float]` vs the `int` `starting_bin`) — leaving as-is, with reasoning.** PEP 484's numeric tower makes `int` acceptable wherever `float` is annotated ("when an argument is annotated as having type float, an argument of type int is acceptable"), so the annotation is already correct as written and `dict[str, float | int]` would be redundant per typing semantics. The `int(params["starting_bin"])` at the setigen call site isn't compensating for annotation looseness — it's documenting that `get_frequency(index=...)` specifically needs the integral value regardless of what the dict nominally carries.

**3. `mode="global"` clearing shell-set env — intentional, now documented.** Added the ops-facing line to `GPUConfig.gpu_thread_mode`'s docstring: the config field is the source of truth and "global" actively clears `TF_GPU_THREAD_MODE`/`COUNT`, exactly the `use_async_allocator` vs `TF_GPU_ALLOCATOR` semantics established directly above it in `setup_gpu_strategy`.

On the non-blocking observations: agreed on all three. `--array-dtype` for `bench_datagen` is deliberately out of scope until the fp16 flag-flip lands (cross-dtype checksums can't gate anything; the wall-time delta of the flip is exactly what the flag-flip PR should measure through the harness). The `array_dtype`-vs-`dtypes` manifest split and the `setdefault` asymmetry in the sweep worker init are both as-designed for the reasons you stated.

Status since the review: the 9-run val-metric A/B (3 seeds × control/fp16/bf16, scaled shape) is running on blpc3 now — results will land on this PR as a comment; per the PR body both flags stay default-off here regardless of its outcome.

@claude please take a second pass over the changes since your review (`5d63f54`).

zachtheyek:
Two updates since the LGTM — a bug found and fixed by the first full-scale run, and the promised val-metric A/B results.

## 1. Production-scale bug in the sweep worker init — fixed in `f391a03`, validated

The first end-to-end runs exposed a bug the CI byte-identity test structurally could not catch: `_sweep_worker_init` pinned `NUMBA_NUM_THREADS=1`, but UMAP calls `numba.set_num_threads(cpu_count)` on its large-N paths and numba hard-errors when asked to grow past the launched pool — every `plot_latent_space_gif` at scale failed non-critically (`Cannot set NUMBA_NUM_THREADS to a different value once the threads have been launched (currently have 1, trying to set 32)`). Small-N unit fits take the brute-force path and never touch the thread API, which is why the pin survived CI.

**Fix**: numba and OMP unpinned in sweep workers (OMP may back numba's threading layer — same trap, different name); BLAS-family pins stay. Shrinking is always legal, so an unpinned launch reproduces the pre-parallel serial semantics exactly. **Validated** by manifest-resuming one failed run at `f391a03`: `vae_plots` retried alone, all 24 GIFs rendered, stage complete, exit 0.

**Timing honesty** (`bbbd901` corrects the docs): the validated sweep took **~93 min at a reduced 60-frame shape with 24 workers** — well below the serial per-combo × 24 bound, far from core-count scaling. 24 concurrent single-threaded UMAP fits/transforms contend for memory bandwidth. The production-scale number, the fit/transform/JIT attribution, and whether a smaller worker cap beats 24 under contention are recorded as open follow-ups; the earlier ~10–15 min expectation is withdrawn.

## 2. Val-metric A/B: 3 seeds × {control, fp16, bf16}, scaled 2×30 shape

| metric | control (s11/s12/s13) | fp16 | bf16 |
|---|---|---|---|
| val_roc_auc | .9987 / .9990 / .9925 | .9982 / .9987 / .9986 | .9981 / .9985 / **lost** |
| test_roc_auc | .9995 / .9979 / .9961 | .9996 / .9986 / .9990 | .9995 / .9987 / **lost** |
| test_recall_at_fpr | .9907 / .9837 / .9449 | .9907 / .9756 / .9619 | .9954 / .9756 / **lost** |
| active_units | 8 / 8 / 8 | **7** / 8 / 8 | **7** / 8 / **lost** |

**fp16 (`round_array_dtype`): parity-leaning.** All three seeds sit inside the control seed spread on every score metric (control's own seed 13 is the batch's most volatile run at .9925 val AUC). One open item before a flip: fp16-s11 measured 7 active units vs control's 8/8/8 — `active_units` is a threshold-crossing count (dim variance vs 0.01), so a dim hovering at the margin flips by luck; the per-dim variance margins need inspection before treating this as signal.

**bf16 (`mixed_precision`): not cleared.** Seeds 11/12 are clean and inside spread, but **seed 13 was the only run in the batch to trip the ECE gate** (winner ECE 0.0634 vs `rf.max_ece` 0.05 — all 8 other runs skipped calibration), triggering the sigmoid calibrator, which compressed every score below the 0.99 screen → degenerate screening (recall 0/0, threshold NaN). One event at n=3 on the batch's most volatile seed is not damning, but it is exactly what this gate exists to catch: bf16 needs more seeds (and a seed-13 repeat with metrics recovered) before any default flip.

**The lost bf16-s13 metrics are a pre-existing master bug, now filed as #289**: the NaN screening stat binds as SQL `NULL`, violates `training_stats.value NOT NULL`, and the failed `executemany` silently drops the entire buffered batch — every RF metric row for the tag. None of this PR's code touches that write path.

**Bottom line**: both flags stay default-off exactly as this PR ships them; no flip is proposed. fp16's flip candidacy is one active-units margin check away; bf16's needs a broader seed sweep after #289 is fixed.

@claude one more pass please, over `f391a03` (the numba/OMP unpin) and `bbbd901` (doc corrections) — the changes since your second review.

zachtheyek:
#290 is merged into this branch verbatim (`ec61dd2`) — the shape-mismatch rephrasing is an improvement over my wording and CI is re-running on the docs-only merge commit.

Closing the review loop here rather than re-tagging: pass 1 produced three notes (two fixed, one rebutted and accepted), pass 2 was a clean LGTM, and pass 3's only finding was a docs phrasing nit that the reviewer itself authored the fix for — a fourth pass would be reviewing its own paragraph, which is the churn condition the review procedure stops on. Everything substantive stands: CI green on every commit, the numba fix validated end-to-end at scale, the A/B verdict published above, both numerics flags shipping default-off.

Ready for maintainer review.

Issue #293, opened 2026-07-29
[SCIENCE]: Calibrate the VAE L1/L2 regularization coefficients before any activation


zachtheyek:
## Background

The conv layers have declared `activity_regularizer=l1(0.001)` and `kernel/bias_regularizer=l2(0.01)` since inception, but they were **dead code**: the custom training loop never consumed `model.losses`, so every model this pipeline has ever trained is effectively unregularized. Surfaced by the #287/#288 optimization audit; activation was attempted on the maintainer's call and **measured harmful at the declared coefficients**, so v1 ships with regularization OFF (`beta_vae.regularization_active = False`) and the objective byte-identical to the historical pipeline.

## The measurement (5-seed reg-on controls vs 3-seed unregularized controls, scaled 2x30 shape, blpc3)

| | val AUC | recall@0.01FPR | active latent dims |
|---|---|---|---|
| unreg (s11-13) | .9925-.9990 | .945-.991 (median .984) | 8 / 8 / 8 |
| reg-on (s11-15) | **.9547**-.9984 | **.725**-.967 (median .954) | 7, 6, 7, **4**, 7 |

- Recall — the primary science metric — dropped on every seed; seed 13 cratered (.94 → .72).
- 1–4 latent dims per seed pushed below the 0.01 active-units variance threshold; seed 14 lost half the latent space (three dims at ~0.002–0.007).
- Training itself was healthy (reg component decays ~240 → ~35-41, no NaN) — the penalties are simply too strong for this model at these coefficients.

## Why the declared coefficients are suspect

- They were never chosen against a working implementation (dead since inception).
- keras activity penalties are **batch-SUM-scaled** (not mean), so the effective activity coefficient scales with per-replica batch size (128 at defaults) — plausibly ~two orders of magnitude stronger than intended.
- Activity penalties also accrue once per regularized forward inside the loss (reconstruction + both clustering branches), further multiplying the effective weight.

## The ready-made harness

[...]

Issue #282, opened 2026-07-26
[FEATURE]: latent representation overhaul — RF variant sweep, deterministic two-pass inference, posterior-collapse diagnostics


zachtheyek:
## Summary

The pipeline currently trains the RF on the **stochastic sampled latent `z`** ([train.py:2429](../blob/master/src/aetherscan/train.py#L2429)), validates on the same sampled `z` ([train.py:2499-2501](../blob/master/src/aetherscan/train.py#L2499)), and runs inference on sampled `z` ([inference.py:427](../blob/master/src/aetherscan/inference.py#L427)) with **no seeding at all** (#279). That combination is self-consistent but has three problems: inference is not reproducible, we never tested whether `z` is even the best representation, and the validation number measures a configuration nobody deliberately chose.

This issue covers the latent-representation work end to end: pick the representation empirically, make inference deterministic and uncertainty-aware, detect posterior collapse, and document the reasoning.

**Prerequisite:** #279 (unified seeding) — every MC draw here must be reproducible, or we reintroduce the exact problem this work exists to fix.

---

## 1. Train all latent-representation variants in one run

Data generation (~302 GB/round) and encoding dominate cost; an RF fit is cheap (`benchmarks/bench_rf.py`). So train **every** variant on the **same generated data and the same train/val split**, and pick the winner empirically.

**Capturing the inputs is free.** [train.py:2429](../blob/master/src/aetherscan/train.py#L2429) is `_, _, concat_z = self.vae.encoder(...)` — it already computes `z_mean` and `z_log_var` and discards them. Keep all three.

Variants (features per cadence, `num_observations=6 × latent_dim=8`):

| # | variant | features | rationale |
|---|---|---|---|
| 1 | `z` (single sample) | 48 | current baseline — keep for comparison |
| 2 | `z_mean` | 48 | deterministic; leading inference candidate |
| 3 | `z_mean + z_log_var` | 96 | full per-dimension uncertainty |
| 4 | `z_mean + per-observation mean log_var` | 54 | ON/OFF-aware aggregate |
| 5 | `z_mean + per-dim mean log_var over obs` | 56 | dimension-wise aggregate |
| 6 | `z_mean + total KL per cadence` | 49 | cheapest aggregate |
| 7 | `z`-augmented (K draws as extra rows), evaluated on `z_mean` | 48 | tests sampling-as-augmentation for a deterministic deployment |
[...]

PR #294, opened 2026-07-29
perf+fix: #289 NaN hardening, fp16 default ON / bf16 off (A/B verdicts), regularization arc → #293, index audit, RF pre-gen, SHAP overlap

### LINKED ISSUE #289: [BUG]: NaN stat value silently drops an entire training_stats flush batch (NaN -> SQL NULL -> NOT NULL violation)
## What happened

During a val-metric A/B batch (scaled 2x30 runs on blpc3, branch `feature/train-perf-optimizations`), one run (`test_20260728_060715`) produced a NaN screening-threshold stat (`largest zero-loss screen on this split: nan` — the run's sigmoid calibrator compressed every score below the 0.99 screen on the small split, so the zero-loss threshold was undefined). When that stat reached the DB writer:

```
aetherscan.db.db | ERROR | Error flushing data buffer: NOT NULL constraint failed: training_stats.value
```

Python's sqlite3 binds `float('nan')` as SQL `NULL`, `training_stats.value` is `NOT NULL`, and the failed `executemany` **dropped the whole buffered batch** — every RF eval metric row for that tag (test_roc_auc, test_recall_at_fpr, ECE/Brier, active_units, all variant_* stats) is missing from the DB, with only an ERROR log to show for it.

## Two defects

1. **No NaN sanitization on the training_stats write path.** `injection_stats` already has `is_finite` semantics for exactly this (non-finite values recorded with `is_finite=0`); the training_stats path has no equivalent, so any NaN metric is a landmine.
2. **One bad row poisons the batch.** The flush is all-or-nothing: a single unbindable row silently discards every other row in the buffer. A per-row fallback (retry the batch row-wise on constraint failure, skipping+logging the offenders) would bound the damage to the bad row.

## Repro material

Logs preserved on blpc3: `/tmp/numerics_ab/run_bf16_13.log` (the failing run), `/tmp/numerics_ab/arms.txt` (the batch). The upstream NaN producer (screening-threshold computation when zero candidates survive a calibrated screen) is arguably a third, separate question — whether that stat should be written as NaN at all or the screen validation should handle the zero-survivor case explicitly.

Found during #287 / PR #288 validation; pre-existing on master (none of the changed code touches this write path).

zachtheyek:
Closes #289

## Summary

Four workstreams, one branch, per maintainer direction: (1) the #289 DB write-path hardening; (2) the fp16/bf16 flag **decisions** from the val-metric A/B campaign — **fp16 default flipped ON (gate passed), bf16 stays off (gate failed on a reproducible pathology)**; (3) the L1/L2 regularization arc — activated on maintainer instruction, **measured harmful, reverted to a default-off observable flag**, calibration filed as #293; (4) the remaining performance items: a full DB index audit, RF-dataset pre-generation on the producer, SHAP/plot overlap, and the GIF-grid trim.

## 1. #289 — NaN stats can no longer vaporize a flush batch

- **v6 migration**: `training_stats.is_finite` (the injection_stats semantics). Non-finite/None values store as 0.0 + `is_finite=0` + a warning; `query_training_stat` filters them by default. The `DEFAULT 1` backfill is exact — a non-finite value could never have been written before (it bound as SQL NULL and the NOT NULL constraint rejected the whole batch — how bf16-seed-13 lost every RF metric in the #288 A/B).
- **Resilient flush**: a failed `executemany` batch falls back to per-row inserts — only the unbindable row(s) drop, with an exact count logged. Before: one bad row silently discarded every row in the flush, across all tables.
- Unit tests: NaN/None storage + filtering + same-batch survival, poisoned-batch fallback, v6 migration backfill.

## 2. Flag decisions (the pre-agreed gate, judged on the unregularized configuration)

Evidence: two batches of seeded scaled-shape runs (2×30, 48000/9600, blpc3), combined at the **distribution level** — a methodology note below explains why same-seed cross-build pairing is invalid.

| arm | seeds | val AUC | recall@0.01FPR | active units | calibration trips |
|---|---|---|---|---|---|
| control | 6 | .9925–.9990 | .9449–.9907 | 6–8 | 0 |
| fp16 | 4 | **≥ .9979 all** | .9619–.9907 (all inside control spread) | 7–8 | 0 |
| bf16 | 7 | .9807–.9991 | **.8432**–.9961 | 6–8 | **2 (both seed 13)** |

- **fp16 → `training.round_array_dtype = "float16"` by default.** Every seed inside the control envelope on every metric. Payoff: round footprint ~294.5 → ~147 GB, gather volume and page-cache working set halved — the lever that matters once two overlapped rounds (~589 GB fp32) exceed the hosts' 503 GB RAM, and it halves the release run's `--keep-round-data` disk. `"float32"` restores the historical input numerics byte-for-byte; `.done` manifests gate every reuse/resume path by dtype; the cli disk-budget guard is dtype-aware.
- **bf16 → stays `False`.** Six of seven seeds clean — but seed 13 reproducibly degrades (recall .8432 / val .9807 vs the .9449 / .9925 control floor), and bf16-seed-13 was the *only* configuration to trip the ECE→calibrator gate, twice, across configs (its first trip is also what exposed #289). A real bf16-specific trajectory pathology; the +1.15× step / ~halved-VRAM upside stays on the table pending an explanation.
- **Gate refinement, recorded in docs**: the "identical active-dim count" criterion is retired — the *controls themselves* flip AU 6–8 across seeds at this shape. The per-dim variance logging added in this PR (see §4) is what shows those flips are marginal-dim wobble, not collapses — and is the criterion's replacement.

## 3. The regularization arc (activate → measure → revert), and #293

The #288 audit's M5 flag is confirmed at code level: the conv layers' declared L1/L2 regularizers were **dead code since inception** — the custom loop never consumed `model.losses`, so every model this pipeline has trained is effectively unregularized. On maintainer instruction they were activated; the 5-seed A/B measured the declared (never-calibrated) coefficients as clearly harmful — recall median .984 → .954 with a worst seed of **.72**, and 1–4 latent dims per seed pushed dark (one seed lost half the latent space). Reverted on maintainer instruction:

- `beta_vae.regularization_active` (default `False`) — the objective is byte-identical to the historical pipeline.
- `reg_loss` is **computed and recorded in every run regardless** (keras materializes the activity penalties anyway — observability costs one `add_n`): new stat rows, a fifth loss-curves panel, and the per-dim variance log line. #293 inherits a ready-made harness plus live data in every future baseline.
- Two crashes/fixes from the arc, both caught by the new blpc3 container **test gate** that now fronts every cluster batch: the FuncGraph stranding (clustering-loss fns inlined — see the bit-compat note) and two test-side arity/determinism fixes.
- **#293** carries the calibration campaign: the declared coefficients are plausibly ~two orders too strong (keras batch-SUM activity scaling × per-forward accrual), with the full measurement as evidence.

## 4. Performance

- **DB index audit** (maintainer mandate; all folded into one unreleased v7 migration; every claim benched with EXPLAIN-verified planner choices, no ANALYZE dependence):
  - `injection_stats`: the stat-scoped index — the end-of-run plot pass (~165 queries) extrapolated to **~6 h of scanning at 367M-row release scale; now ~8 min** (14.4× per query at 2M rows). Write cost disclosed: bulk insert throughput −61%, still ~235× over the lane's real demand. The trailing-`timestamp` shape is deliberate: a `round_number`-trailing variant is *never chosen* by the planner without `sqlite_stat1` (measured; recorded so it isn't "fixed" back).
  - `latent_snapshots`: reshaped to the GIF fetch key — **209 → 3.1 ms per frame fetch (67×; ~1.46 h → ~1.6 s at 100M-row release scale)**, dashboard latest-capture 595 → ~0 ms, write cost +0.9% (both shapes append at the B-tree right edge). Old index dropped.
  - **Measured rejection, recorded**: a training_stats reshape wins one minor query but makes the dominant loss-curves query 1.8× *slower* and costs −19.5% insert throughput. Every other table: left alone with the arithmetic shown. `bench_db_index_shapes.py` + `bench_injection_index.py` reproduce all numbers.
- **RF-dataset pre-generation on the producer**: the producer now generates the RF set during the last round's training (byte-identical: same seed-derivation sentinel; pinned by a test driving the real spawn protocol against in-process generation). Await → validate → unchanged in-process fallback; all eight lifecycle paths (success/error/no-producer/resume×2/crash/skips/teardown) pinned with an ordered event log. Every A/B matrix run exercised this path end-to-end at scale.
- **SHAP overlap + single pool**: rf_plots renders the 5 non-SHAP diagnostics while SHAP computes on a background thread (wall saving ≈ min(SHAP, plots)); one forkserver pool serves all three passes with per-pass-family explainer caching and ~4 chunks/worker straggler smoothing. Bitwise contract pinned (pooled ≡ single-process per pass); failure semantics preserved exactly, with one intentional delta: a failed SHAP compute no longer re-attempts the ~1.4 h computation five times to record five failures.
- **GIF sweep default 24 → 18 combos** (`n_neighbors` 30 dropped as redundant between 15 and 50 — maintainer request).

## Validation

- Container test gates fronted every cluster batch (the gate caught two real bugs before they cost matrix hours); full unit suite green on the final tip: **960 passed**, 5 deselected, 6m03s on the final tip.
- 25 scaled A/B runs across three matrices (5 reg-on controls, 9-run final decision matrix, plus the #288-era batch), every completed run exit-0 on the final code.
- Tip mini-run (fp16 as default, producer RF pre-gen, SHAP overlap, single-combo GIF): **clean end-to-end, exit 0** (2 rounds × 5 epochs at 48000/9600). Exercised in one run: fp16 as the default (dtype-recorded manifests, including retry-reuse through the dtype gates), the producer-delivered RF dataset (**0.0 s wait** at rf_train; reused across retries), the SHAP/plot overlap, and all 10 RF diagnostics. The mini-run also caught the last bug in this PR — the hardcoded 48-name SHAP feature list breaking on a 54-feature variant winner (`52ccecf`, pre-existing #282 fallout that no prior run had tripped) — and surfaced one pre-existing CLI validation gap worth a follow-up issue: `--num-samples-rf 1600` passes parse-time validation despite conflicting with the effective batch on a 5-GPU host (the runtime backstop catches it, as designed).

## Notes for reviewers

- **Bit-compat break, loud and deliberate**: inlining the clustering-loss fns (the FuncGraph fix) changed float summation order — same-seed results across builds differ slightly; within-build determinism is preserved and pinned by the existing test suite. This is why the A/B evidence combines batches at the distribution level only.
- Commit history carries three cross-workstream sweeps (concurrent agents sharing files: `4d931ae`, `7b45c7c`, and the docs half of the SHAP change) — content is correct and complete; the messages note the sweeps.
- blpc3's disposable test DB accumulated pre-reshape v7 index names during the matrices; it gets dropped in the post-merge teardown. No production DB is affected (bla0 migrates v5 → v7 cleanly on its restart).
- The GIF-sweep worker-cap benchmark (bandwidth-contention question) is recorded in the handoff as a production-scale follow-up measurement.

### DIFF
--- a/benchmarks/bench_db_index_shapes.py
+++ b/benchmarks/bench_db_index_shapes.py
@@ -1,0 +1,435 @@
+#!/usr/bin/env python3
+"""
+Micro-benchmark: the schema-v7 index reshapes on training_stats and latent_snapshots
+(companion to bench_injection_index.py, which covers the injection_stats side of v7).
+
+For each table it synthesizes a production-shaped database twice in a scratch directory —
+once with the old (tag, timestamp, ...) filter index, once with the v7 equality-first
+replacement — replicating db.py's schema, pragmas (WAL + synchronous=NORMAL) and each
+table's real write pattern (training_stats: writer-thread flush batches of 5,000 rows;
+latent_snapshots: one 3,840-row executemany per snapshot capture). It then times every
+production query shape observed in the codebase against both variants, with an EXPLAIN
+QUERY PLAN line per shape so the chosen index is visible. No ANALYZE is ever run — db.py
+never runs it, so a plan that needs sqlite_stat1 is a plan production never gets.
+
+Shapes covered (and their call sites):
+- training_stats: the run-wide loss-curve fetch (train.py plot_beta_vae_loss_curves),
+  the stat-scoped snr_range_floor/ceil fetch (train.py _get_snr_by_round), the kl_dim_*
+  IN fetch (train.py plot_posterior_collapse), and the dashboard's stat_name IN + ORDER BY
+  poll (dashboard.py load_training_stats).
+- latent_snapshots: the per-frame capture fetch (train.py plot_latent_space_gif — up to
+  latent_viz_gif_max_frames=500 of these per run), the dashboard's latest-capture lookup
+  (ORDER BY round/epoch/step DESC LIMIT 1), and the DISTINCT key enumeration
+  (query_latent_snapshot_keys) as a regression check — it must stay a bounded tag scan.
+
+Verdict baked into schema v7 (measured on an M-series Mac, sqlite 3.50): the
+latent_snapshots reshape ships — 67x per frame fetch at 2M rows (linear in partition size:
+~1.5 h -> ~2 s per GIF pass at the 100M-row release scale), the latest-capture lookup
+becomes a backward index walk (595 ms -> ~0 ms), keys_distinct unchanged, write cost
+within noise (both shapes append at the b-tree right edge — round/epoch/step grow
+monotonically like timestamp). The training_stats reshape was REJECTED on this bench's
+numbers: it won only the single-stat shape (3.5x on ~9 ms) while regressing the dominant
+run-window loss-curve fetch 1.8x (stat-sorted index order scatters table-row lookups that
+timestamp order visits sequentially), the kl_dims/dashboard IN shapes ~1.2x, and insert
+throughput ~20% — on a table whose tag partitions stay ~58k rows.
+
+    python benchmarks/bench_db_index_shapes.py [--training-tags 18] [--latent-captures 520]
+
+Prints per-shape ms and rows/s per variant plus a release-scale extrapolation of the GIF
+frame pass, and writes a JSON result to benchmarks/results/ (or --output).
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sqlite3
+import tempfile
+import time
+
+import numpy as np
+from _common import time_repeats, write_result
+
+BASE_TIMESTAMP = 1_700_000_000.0
+FLUSH_BATCH_ROWS = 5_000  # config.db.write_buffer_max_size — the writer's flush batch
+METADATA_JSON = json.dumps(
+    {"machine_name": "bench", "user_name": "bench", "ip_address": "0.0.0.0", "process_id": 0},
+    sort_keys=True,
+)
+
+# --- training_stats: ~29 rows/epoch (21 scalars + latent_dim=8 kl_dim rows), 20 rounds x
+# --- 100 epochs per tag; a campaign db accumulates one such partition per run tag ---
+TRAINING_STATS = (
+    ["total_loss", "reconstruction_loss", "kl_loss", "true_loss", "false_loss"]
+    + ["val_total_loss", "val_reconstruction_loss", "val_kl_loss", "val_true_loss"]
+    + ["val_false_loss", "gradient_norm_mean", "gradient_norm_max", "gradient_norm_std"]
+    + ["clipping_rate", "learning_rate", "epoch_duration", "steps_per_epoch"]
+    + ["train_samples", "snr_range_floor", "snr_range_ceil", "beta"]
+    + [f"kl_dim_{d:02d}" for d in range(8)]
+)
+KL_DIM_STATS = [s for s in TRAINING_STATS if s.startswith("kl_dim_")]
+DASHBOARD_STATS = TRAINING_STATS[:5] + TRAINING_STATS[5:10]  # losses + val_ variants
+ROUNDS, EPOCHS = 20, 100
+
+TRAINING_TABLE_SQL = """
+    CREATE TABLE IF NOT EXISTS training_stats (
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        timestamp REAL NOT NULL,
+        model_name TEXT NOT NULL,
+        stat_name TEXT NOT NULL,
+        value REAL NOT NULL,
+        round_number INTEGER,
+        epoch_number INTEGER,
+        tag TEXT,
+        metadata TEXT,
+        superseded INTEGER DEFAULT 0,
+        is_finite INTEGER DEFAULT 1
+    )
+"""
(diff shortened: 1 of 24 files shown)


zachtheyek:
All three inline notes accepted and fixed in `fc0216b` — no rebuttals this round:

**1. `_executemany_resilient` mid-batch duplicates — real, fixed with SAVEPOINT.** You're right that `executemany` steps rows into the open transaction, so a mid-batch failure left the pre-offender rows in place and the per-row retry duplicated them — and right that the existing test hid it by poisoning index 0. The batch attempt now runs inside `SAVEPOINT resilient_batch` with `ROLLBACK TO` before the per-row pass, and a new test pins the mid-batch case (good, poison, good → exactly one copy of each good row). The docstring's "only the offending row(s) drop" claim is now actually true.

**2. Bit-compat warning in-source — added.** The clustering-fn inlining comment now carries the reproducibility consequence directly (summation order changed; same-seed cross-build differs; within-build determinism pinned; distribution-level comparisons only), so nobody debugging reproducibility needs PR archaeology.

**3. Explainer cache-key guard — added.** `_get_explainer` validates the pass kind itself rather than collapsing unknowns into the raw slot, closing the in-process path that bypasses `_validate_pass`.

CI is re-running on `fc0216b`. @claude one more pass over the changes since your review, please.

PR #299, opened 2026-07-30
perf(inference): corrected I1-I9 plan — 2.1x end-to-end, byte-identical preprocessing, determinism by default

### LINKED ISSUE #298: [FEATURE]: Inference pipeline performance optimizations (corrected I1-I9 plan + fresh-eyes audit findings)
**Is your feature request related to a problem? Please describe.**

CSV-catalog inference over real `/datag` `.h5` cadences is wall-clock dominated by avoidable overheads, on both of its two operating regimes:

- **Fresh runs** (first pass over a catalog): the prefetch thread's preprocessing is the critical path (~2–6 min/cadence). The dominant cost is a pathology: `_extract_stamps_worker` opens each obs `.h5` with h5py's **default 1 MiB chunk cache**, which cannot hold a single ~4 MiB decompressed bitshuffle chunk — so every stamp read re-decompresses ~16 full chunks (~64 MiB) to extract 256 KiB, and the start-sort at `preprocessing.py:1699` (whose in-code comment *counts on* chunk-cache reuse) buys nothing. On top of that, extraction dispatches 36 near-equal tasks onto 32 workers (~44% worker-idle tail in the measured bottleneck stage), and prefetch depth 1 serializes each cadence's disk-bound energy detection behind the previous cadence's decompression-bound extraction.
- **Cached-stamp reruns** (same-tag retries, threshold sweeps, re-inference with new weights — and every Phase-3 iteration loop): the main thread becomes the critical path. `run_inference` rebuilds, re-distributes, and re-`iter()`s a fresh `from_generator` dataset **per cadence** (~9.1 s/cadence of iterator churn alone, feeding ~2.8 GB/s against ~38 GB/s of demand — the same disease #276 fixed for training); every cadence is padded up to a full 10,240-snippet global batch (a 100-stamp cadence encodes 10,240 rows, and each replica runs a 12,288-obs forward — above the documented ≳8192-obs int32 abort observed in `bench_gpu`, never yet exercised by a real 5-GPU streaming run); the MC cascade makes 32 sequential full-forest `predict_proba` calls per surviving cadence (dispatch-dominated) plus 32 more at reference-cloud finalize; full `gc.collect()` runs up to ~6× per cadence including once per encode step (~0.33 s each measured); and the streaming load path is a per-stamp Python loop with a restack and 5 redundant validation passes. Fresh-tag reruns also re-pay the entire preprocessing cost because the stamp cache directory embeds `save_tag`.

Two independent audits (the session-4 12-agent pass that produced the I1–I9 plan, and a fresh-eyes 15-agent adversarially-verified pass on 2026-07-30 that re-derived the same headline items without seeing that plan and adjudicated every disagreement) converge on the ranked, regime-annotated plan below. Estimated impact: **~40–60% off a fresh 100-cadence run** (I1 alone), with the rerun-regime encode phase dropping ~20–25 min → ~3–5 min.

**Describe the solution you'd like**

Performance-only changes — detection/scoring science untouched. Every item carries an explicit numerics class (`none` = byte-identical, gated by `np.array_equal`/byte-compare; `bit_compat_break_only` = same science different low bits, gated by candidate-set + `mc_mean`/`mc_std` equality on a real-cadence A/B):

1. **I1 — HDF5 chunk-cache fix** in stamp extraction (+ the PFB residual-check reads): open with `rdcc_nbytes` sized at runtime from `dset.chunks` (graceful on other layouts), large-prime `rdcc_nslots`, `rdcc_w0=0`. Numerics: none. Gated by an `h5ls` chunk-shape diagnostic on real `/datag` files.
2. **N1 — extraction task granularity**: ~4 tasks/worker (`ceil(4*n_processes/len(h5_paths))` per obs) + `imap_unordered` (workers write disjoint absolute rows; byte-identical). Numerics: none. Also the hedge if I1's chunk-shape assumption fails.
3. **N2 — prefetch depth 2** (`inference.prefetch_depth`, default kept at 1 until the post-I1 on-cluster A/B decides the flip): overlaps disk-bound ED with decompression-bound extraction; verified safe against pool-lifecycle, seeding (#279 keys on catalog index), and manifest-ordering constraints. Numerics: none.
4. **I2⊕I4 — encode-path rework, option (b)**: drop `tf.data` on the inference encode path; cached per-replica concrete functions fed numpy slices via `strategy.experimental_distribute_values_from_function`; batch geometry chosen there (bucketed per-replica sizes, `per_replica_batch_size` default 2048 → 256 snippets, bounded padding). Numerics: bit_compat_break_only. Includes the Phase-3 abort-reachability probe (one pipeline-shaped 12,288-obs forward) so the int32-abort risk is established or retired before the real `/datag` run.
5. **I8 — batch the MC cascade + reference-cloud scoring** into one `predict_proba` per pass (draws still generated sequentially from the same RNG → stream order unchanged; equality validated at `n_jobs=1`). Numerics: none.
6. **I7 — remove the gc.collect() storm** (`inference.py:460/468/694`, `main.py:501`, per-chunk collects in `load_inference_data`); keep exactly the finally-block collect at `inference.py:600`. Gate: RSS-flatness over ≥20 cadences. Numerics: none.
7. **I5 — vectorize the streaming lognorm/load path** (batched keepdims reductions into a preallocated float32 output; single-pass validation) and overlap the load with the prefetch task. Numerics: none (`np.array_equal` gate). Sized honestly: rerun-regime win.
8. **I3 — fingerprint-keyed preprocessing cache, default flip** (maintainer-approved): default stamp dir becomes `preprocessed/<csv_stem>_ed<fingerprint12>/` keyed on a **denylist** ED fingerprint (excluding model/scoring/viz/retry keys, so unknown new keys over-invalidate — an allowlist would silently reuse wrong stamps on e.g. a `coarse_channel_width` change), with a provenance guard (sidecar `h5_paths` validated against the group) and per-run-unique tmp names (closes the concurrent-run tmp-deletion race). Numerics: none. Reverses the documented tag-isolation trade (deliberate; scores stay tag-scoped in the DB).
9. **I6 leftovers**: single ordered file-major imap across the 3 ON files (removes 2 pool drain tails; byte-identical assembly), PFB residual check moved onto the pool (per-cadence warning semantics preserved — the check's documented cross-file consistency diagnostic must not be memoized away).
10. **I9 — viz tail**: single background FIFO Slack uploader (joined before teardown) + process-parallel candidate-figure rendering via the `latent_gif.py` forkserver/empty-preload pattern (the figure-render path must be extracted TF-free first — `inference_viz` transitively imports TF via `aetherscan.models`, and bare fork of the CUDA-initialized parent is not acceptable).
11. **Riders** (byte-identical or output-identical cleanups on touched files): compact metadata JSON separators (~0.5–1.5 s/cadence on the prefetch critical path), per-worker h5 handle reuse for ED (2,859 opens/cadence today), memoized PFB sidecar verification, `np.argsort`-based hit dedup, `_GLOBAL_CHUNK_DATA` release (~6–10 GB pinned RSS), ED-pool + first-prefetch start before model load (hides the 10–60 s model load), `_write_inference_results` over `np.nonzero`, `prepare_latent_features` as a reshape, dataset `.repeat()` removal (dies with I2), single `db.flush()` in the viz suite, candidate-gallery column trim.

**Validation** (each item ships with unit tests; on-cluster steps run on blpc3 after the current training run completes, against its `train_20260729_125502` weights): byte-compare gates for all numerics-none items; candidate-set + `mc_mean`/`mc_std` equality for I2⊕I4; one clean pre-optimization baseline `/datag` run first (doubles as the missing #183 inference numbers); stage-span A/Bs from `pipeline_stages` for each landed item.

**Describe alternatives you've considered**

- *Per-item alternatives rejected by the audits*: keeping `tf.data` with a blocking batch-slice generator (subtle stale-prefetch + rebatch-shape hazards vs. option (b)); a standalone `per_replica_batch_size` config flip (leaves the dominant iterator-rebuild cost); an ED-key **allowlist** fingerprint for I3 (unsafe — silently reuses wrong stamps when a non-enumerated key changes); memoizing the PFB check once per geometry (destroys its documented cross-file consistency diagnostic); parallel candidate rendering via bare `multiprocessing.Pool` fork (forks a CUDA-initialized parent).
- *Deprioritized as sub-1% after adversarial verification*: ED-worker CPU micro-optimizations (disk-bound stage), metadata list→array conversion at viz time, reference-cloud-only batching without the per-cadence half.

**Additional context**

Fresh-eyes audit: 5 independent domain readers (ED core / stamps+load / encode+classify / orchestration / viz tail) × 2 adversarial verifiers each (refute lens + Amdahl/science lens), 21 findings, every cited site verified against `master` @ `cb1e764`. Wall-model consensus for a 100-cadence fresh run: ~7–12 h, of which extraction decompression ~50–60% and ED reads ~30–40%; GPU encode + RF are fully hidden under prefetch (GPUs >95% idle by design pre-I2). The two-regime split (fresh vs cached-stamp) is what re-orders the original I1–I9 ranking: I7/I8/I4/I5 are ~0% on fresh runs (hidden under the prefetch shadow) and real on reruns; I1/N1/N2 are the fresh-run levers.

**Discussion links**

Maintainer-scoped and approved in the working session of 2026-07-30 (decisions: I3 default flip with denylist fingerprint + provenance guard + unique tmp names; I2⊕I4 via option (b); full corrected plan in one PR; full I9 with the fork-safety treatment). Follows on from the merged training-side optimization arcs #283/#288/#294.

zachtheyek:
Closes #298.

Implements the full corrected inference-optimization plan — the recovered session-4 I1–I9 items, re-audited by a fresh-eyes 15-agent adversarial pass (5 domain readers × refute + Amdahl/science verifiers), with every surviving item validated on blpc3 against real `/datag` data and the `train_20260729_125502` weights. Detection/scoring science is untouched: every change is either provably byte-identical (gated by `np.array_equal`/byte-compares) or bounded bit-compat drift (gated by candidate-set equality under deterministic ops — see the numerics section).

## Headline numbers (blpc3, 2-cadence RFI-dense L-band subset, 73k + 21k stamps, same weights/catalog/seed)

| | master | this branch | |
|---|---|---|---|
| End-to-end wall | 22.5 min | **10.5 min** | 2.1× |
| Stamp extraction (per cadence) | 446 s | **189 s** | 2.4× (I1+N1) |
| Energy detection (per cadence) | 115 s | **28 s** | 4.1× (I6 + rider; page-cache warmth inflates this somewhat — optimized ran second) |
| RF two-pass cascade | 8.4 s | **3.0 s** | 2.8× (I8) |
| Reference-cloud finalize | 10.4 s | **1.5 s** | 6.9× (I8) |
| load+log-norm | 21.2 s | **13.7 s**, moved off the GPU thread | I5 |
| Viz suite | 56 s | **42 s** | I9 |
| Cached-stamp rerun (same ED config) | n/a (re-preprocesses) | **~4 min** | I3 |
| **Stamps + metadata** | — | **byte-identical to master** | the numerics-none gate |

Full-catalog projection: the 8 cache-cold A/B cadences averaged ~13.5 min each (depth 2) vs ~5 min for the warmer subset — so all 6,093 cadences project to roughly **3–8 weeks** depending on per-band RFI density and cache behavior (master would be ~2–3× that). ⚠ Disk, not compute, is the binding constraint (stamps ≈ 30–90 TB vs 2.9 TB free on /datax) — automatic stamp-cache pruning is maintainer-approved as a follow-up PR (see HANDOFF).

## What changed (commit per item)

- **I1 — HDF5 chunk cache** (`b06fc9d`): `_extract_stamps_worker` (and the PFB residual-check reads) open with `rdcc_nbytes` sized at runtime from `dset.chunks`. h5py's 1 MiB default can't hold one decompressed BL chunk — confirmed on real `/datag` files: `(1,1,1048576)` bitshuffle float32 = 4 MiB/chunk — so every stamp re-decompressed its ~16-chunk stripe and the start-sort's documented cache-reuse rationale never held. Byte-identical; degrades to defaults on other layouts.
- **N1 — extraction task granularity** (`dd619d8`): ~4 tasks/worker + `imap_unordered` (was 36 near-equal tasks on 32 workers = 2-wave makespan, ~44% worker idle). Rows land at absolute indices → byte-identical under any completion order.
- **I7 — gc.collect storm** (`5394b3a`): removed up to ~6 full collections/cadence (incl. one per encode step, ~0.33 s each with TF loaded, GIL held); kept exactly the one per-call collect in `run_inference`'s finally. RSS stayed flat across all validation runs.
- **I8 — batched MC scoring** (`b792e30`): one stacked `predict_proba` per pass (32 sequential full-forest calls before, per surviving cadence + at reference-cloud finalize). Draws still generated sequentially from the same RNG → #279 stream order byte-identical; bit-identity pinned by unit tests at `n_jobs=1` against a verbatim reimplementation of the old loop.
- **I5 — vectorized streaming load** (`4a35adf`): the per-stamp Python log-norm loop + restack becomes one vectorized pass per chunk into float32 blocks (bit-identical per element — same scalar casts, exact order-independent reductions; test-pinned against `_lognorm_worker` in fp32+fp64), plus the `_GLOBAL_CHUNK_DATA` release (~6–10 GB no longer pinned between cadences).
- **N2 — prefetch depth + load overlap** (`c724ab2`): `inference.prefetch_depth` / `--prefetch-depth`; each prefetch future now preprocesses AND loads/log-norms its cadence; models load after the first prefetch (the 10–60 s load hides under first-cadence ED). Futures consumed strictly in catalog order → results identical at any depth (fingerprint-denylisted). Depth A/B numbers below.
- **I6 — fused ED + pooled PFB check** (`0600653`): one ordered file-major imap over all 3 ON files (byte-identical hit assembly; per-file histograms recovered from task position) instead of three drained imaps; the PFB residual check's 27 band reads ride the pool via `map_async` (bit-identical ratios — the response ships as its verified sidecar path; the cross-file consistency warning contract is unchanged).
- **I2⊕I4 — encode from numpy slices** (`043a0d9`): `_distributed_encode` feeds replicas via `experimental_distribute_values_from_function` — no per-cadence tf.data build/distribute/`iter()` (~9.1 s/cadence of iterator churn), no per-element generator, no full-array pad copy. Bucketed power-of-two batch geometry (floor 16, cap `per_replica_batch_size`, default 2048→**256 snippets** = 1,536-obs forwards near the measured encode peak); padding bounded under one bucketed batch and never leaves the encode. Dead `InfDataHolder`/`prepare_distributed_inf_dataset` removed. (Bench note: master's 12,288-obs forwards did *not* trip the documented int32 abort on Blackwell — the cliff appears bench-specific; the new geometry stays far from it regardless.)
- **I3 — ED-fingerprint-scoped stamp cache** (`403b767`, + `6a730ac`): default stamp dir is `preprocessed/<csv_stem>_ed<hash12>/`, keyed on `preprocessing_config_fingerprint` — a fail-safe **denylist** hash (an ED-key allowlist would silently reuse wrong stamps when an unlisted key changes). Runs sharing an ED config share stamps; any ED change lands in a fresh dir by construction. Shipped with a resume provenance guard (sidecar `h5_paths` + recorded `ed_config_fingerprint`), per-run-unique tmp names (closes the concurrent-run clobber race; abandoned tmps age-swept, legacy fixed name included), and compact metadata JSON (the `indent=2` pretty-print cost ~0.5–1.5 s/cadence on the prefetch critical path). Maintainer-approved reversal of the documented tag-isolation trade; scores stay tag-scoped in the DB.
- **I9 — viz tail** (`a3c70aa`): single-worker FIFO async Slack uploader (figure order + API rate preserved; drained before teardown) + per-candidate figures rendered across a forkserver pool with empty preload in the new TF-free `candidate_figures.py` (`inference_viz` imports TF transitively — a bare fork of the CUDA-initialized parent is exactly what the forkserver avoids). Per-figure failures degrade the suite (`_viz_safe` contract).
- **Riders** (`0aadfa7` + in-item commits): per-worker h5 handle LRU for ED (pool workers only — ~2,859 open/close cycles per cadence removed), stable-argsort hit dedup (provably the identical permutation), `np.nonzero` positives iteration, `.repeat()`/generator machinery gone with tf.data.
- **Saved-config hygiene** (`70c6b8a`): `apply_saved_config` no longer layers host-tuning knobs (`per_replica_batch_size`, `prefetch_depth`) — a pre-#298 training config would otherwise silently re-impose 2048 on every future inference run.
- **Determinism by default** (`bcee631`, maintainer decision): `tf_deterministic_ops` → True (root seed already 11). Opt-outs are explicit CLI acts: `--no-tf-deterministic-ops`, and the new `--unseeded` (mutually exclusive with `--seed`, validated). The `reproducibility` section joins the saved-config skip-set so a pre-#298 config can't silently disable determinism via `--config-path`.
- **Frequency-labeled snippet axes** (`c0246a7`, maintainer request): every cadence-snippet waterfall (per-candidate, candidate gallery, stamp gallery) labels its x-axis with the stamp's frequency span (from the metadata sidecar; best-effort, unlabeled on legacy sidecars).
- **Docs**: full suite sync (`e7ca80f` + in-item), README CLI blocks regenerated.

## Numerics & the determinism finding

The only numerics class in play beyond byte-identical is the encode path's batch-geometry change (bit_compat_break_only). Measured on the subset:

- Score drift on shared candidates: max |Δ| ≈ 2.3×10⁻³ = 1–2 tree-votes of the 1000-tree forest — bit-level latent drift quantized through exact RF split thresholds.
- **Under `--tf-deterministic-ops` (now the default): master and this branch produce IDENTICAL candidate sets** (det-pair runs `inf_20260730_093804` vs `inf_20260730_100142`: 0 flips, 10/10 shared).
- Without the flag, *two identical runs of the same code* already flip 2–3 near-threshold candidates (measured: `nd1` vs `nd2`, 3 flips) — pre-existing cuDNN autotune noise, present on master, not introduced here. Two flagged runs were bit-identical (max Δ = 0.0). This measurement is what motivated the determinism-by-default decision.

## Validation summary

- CI: full unit suite green on 3.10/3.11/3.12 at every push (~1,040 tests; ~60 new).
- blpc3 NGC container: 1036 passed on the branch.
- Real-data A/B on `/datag` (all tags in the DB for auditability): stamps + metadata byte-identical; per-stage spans in `pipeline_stages`; equality/repro harnesses preserved on blpc3 (`/tmp/compare_inference_runs.py`, `/tmp/repro_check.py`, `/tmp/detpair_check.py`).
- Prefetch-depth A/B (8 fresh cache-cold cadences, isolated scratch dirs, deterministic ops): depth 2 = 1h48 vs depth 1 = 2h10 end-to-end (**~16% lower wall**), identical candidate sets (21 = 21) → default flipped to 2 (`e5ab9c0`). Per-stage sums rise under depth 2 (concurrent cadences share disk/CPU) while wall falls — the overlap wins.
- Integration gates against `AETHERSCAN_SMOKE_MODEL_TAG=train_20260729_125502`: **both pass** (`2 passed` in 4m33s). Noteworthy: **both gates were latently broken before this PR** and are repaired here (test-only): the monotonicity gate fed the encoder raw-width (4096) backgrounds — a shape error on any model, on master too — and now downsamples the plate to the persisted encoder's own input width before injection (mirroring `load_train_data`); the inference smoke passed a bare-datetime `--save-tag`, rejected since the #272 tag refactor, and now passes the `inf` prefix and recovers the run's self-stamped tag from its output. It also accepts the saved config from `output_path` (where training actually writes it) in addition to the legacy `model_path` location. The 5-round test model passes the SNR→confidence monotonicity gate. The hours-long train smoke is deliberately deferred to the maintainer's Phase-3 session (training code is untouched by this PR; the shared CLI/config surface is covered by the unit suites).

## Follow-ups filed in HANDOFF (not in this PR, maintainer-directed)

1. **Automatic stamp-cache slicing/pruning** (approved, default-on, own PR) — required before any full-catalog run (disk math above).
2. **Saved-config loading rethink**: `checkpoint` is already never layered (a stale save_tag cannot leak — test-pinned) and #298 excluded host/reproducibility knobs, but the apply-everything-with-skip-lists design is backwards: paths (`data_path`/`model_path`/`output_path`), `gpu`, `manager`, `db`/`logger`/`monitor` still layer from the training host's config. Proposal (allowlist inversion) + open questions recorded in HANDOFF for maintainer scoping.


zachtheyek:
Thanks for the thorough first pass. All four concerns and both nits worked individually; fixes landed in `8e91bac` (CI green ×3.10/3.11/3.12).

## Addressed

**1. Legacy-loader thread contract** — you were right, and it was worse than a missing assert: the legacy full-width sequential branch *staged its chunk in `_GLOBAL_CHUNK_DATA`*, which under `prefetch_depth > 1` with `store_downsampled_stamps=False` is a real cross-thread race, not just an unenforced contract. Rather than asserting `already_downsampled` (which would fail-loud but leave the class of bug in place), the fix removes the module global from the threaded path entirely: the downsample core is extracted as `_downsample_cadence` (a pure function; the pool worker now delegates to it), the sequential branch passes rows to it directly, and the `parallel=False` THREAD CONTRACT is spelled out in `load_inference_data`'s docstring — module-global staging is now a pool-worker-only mechanism by construction.

**3. `_AsyncUploader` timed-out drain** — implemented per your suggestion, with one refinement: each worker thread captures *its own* stop event at start, and `drain()` re-arms a fresh event after signalling — so a timed-out predecessor keeps honoring its signal (discard backlog, exit after the in-flight call) while any successor thread starts clean.

**4. `_flattened_edge_mid_ratio_banded` bound** — `hi` is now clamped to the channel width, with a comment noting the clamp is defensive documentation (the center+3 widening cannot reach the channel end at any real width).

**Nit 2 (outer `del cadence_data`)** — rationale comment added: the callee's `del` only clears its own reference, and with `prefetch_depth` cadences already loading behind the current one, holding the array until the next iteration's rebind would stack an extra cadence-sized array on peak RAM.

## Rebutted (code left as-is, reasoning below)

**2. `_chunk_cache_kwargs` duplicate open** — measured-negligible by construction: the extra open is metadata-only (no data pages touched; ~1 ms on /datax), paid once per extraction *task*, and N1's granularity puts that at ~130–190 tasks/cadence → ≤0.2 s per cadence, inside pool workers, against the multi-minute stage the rdcc sizing accelerates. A path-keyed cache would add cross-worker state (or parent-side plumbing through task args) for a win below measurement noise — taking your own "call it out only if a profile shows it" disposition and leaving it.

**Nit 1 (stale `NOTE` beside `per_replica_batch_size`)** — already gone in this PR: the I4 commit (`043a0d9`) replaced the entire `per_replica_batch_size: int = 2048  # NOTE: come back to this later` line with the documented 256 default and its rationale block; current `config.py` has no marker on that field. The remaining `NOTE:` markers in `config.py` sit on unrelated fields (`num_samples_rf`, cadence grouping, overlap search) and predate this PR — left per the surgical-changes rule.

Ready for a second pass: @claude

PR #305, opened 2026-07-31
Round-2 inference optimizations: catalog-scale survivability, stamp-cache pruning, saved-config allowlist

### LINKED ISSUE #297: [BUG]: CLI validation misses num_samples_rf vs effective_batch_size conflict
Found during #294's final validation: `--num-samples-rf 1600` passes parse-time validation on a 5-GPU host despite conflicting with the default `effective_batch_size=7680` — the RF stage's train split (1600 × 0.8 = 1280 rows) trims to zero and the run dies at rf_train with the runtime backstop error `train_steps < 1: n_train_trimmed (0) must be >= effective_batch_size (7680)`, after the (now-fast, but still wasteful) beta-VAE rounds have already trained.

The cross-param validator (`collect_validation_errors` / `_check_cross_constraints`, #255) checks the beta-VAE sample counts against batch divisibility but apparently not the RF path's `num_samples_rf × train_val_split >= effective_batch_size` requirement (and divisibility thereof). Add the check + a suggested-fix line mirroring the existing cross-param error format, and a matrix test in `test_cli_validation.py`.

Severity: low (runtime backstop catches it, as designed) but the failure costs a full VAE training phase before it fires.

### LINKED ISSUE #301: [PERFORMANCE]: Round-2 inference optimizations from the post-#299 fresh-eyes audit
**Context.** Maintainer-directed follow-up to #298/#299 (interactive session, 2026-07-31): a 12-agent fresh-eyes audit of the post-#299 inference pipeline, grounded in the actual `pipeline_stages` spans of the #298 validation runs on blpc3, surfaced a second round of verified optimizations. Ground truth: in the fresh regime the main thread + 5 GPUs idle **91.5%** of wall waiting on preprocessing; a measured **777 s** un-spanned cold-start is the `Database.__init__` `COUNT(*)` storm (7 tables incl. 190M injection_stats rows, 80 GB DB); in the cached regime **viz is 73% of wall** (114 s of it one figure's cold h5 reads); the end-of-run viz suite holds every cadence's parsed metadata JSON (OOM-class at 6,093 cadences) and the preproc-funnel figure hard-fails past 242 cadences (Agg 65,536-px limit, exception swallowed).

**Scope (all numerics-none unless stated, each gated per the #298 validation protocol):**
- DB: drop the synchronous `get_db_stats()` COUNT storm from `Database.__init__` (measured 777 s/cold launch, re-paid every relaunch); add the missing `(tag, npy_path)` partial index for `mark_superseded` on `inference_results` (schema v8).
- Viz, catalog-scale survivability: stream-reduce instead of retaining all metadata; pre-binned hit-frequency histograms in new sidecars (old sidecars stay readable); width-capped preproc funnel; per-figure sub-spans under `inference.viz`; viz-scope knob for resumed passes; memoized candidate sidecar reads; persist decimated bandpass envelopes at ED time (kills the 114 s cold-h5 viz reads); bounded monitor teardown-plot query.
- Preprocessing: remove the per-task whole-mapping `msync` in `_extract_stamps_worker`; hoist `_chunk_cache_kwargs` to the parent; coalesce overlap-triplet stamp reads (byte-identical); single-pass load validation + no >50k-stamp concatenate copy.
- RF/scoring: `prepare_latent_features` → reshape (byte-equivalent, shared with training, pinned by test); viz-collector budget check before feature build; `_batched_mc_scores` transient halved; predict-time `n_jobs` restored to config control after `joblib.load`.
- Encode (attempt, det-pair-gated): overlap the serialized convert→dispatch→harvest step loop in `_distributed_encode` (runs at ~2% of benched encode throughput today); rider: encode-only submodel skipping the discarded `Sampling` draw.
- Logging: ED progress → milestone lines (25/50/75/100% per ON file); the per-coarse-channel lines are a measured 62% of all log volume, all Slack-bound (~594k lines over a catalog run).
- On-cluster A/Bs (blpc3, data-gated defaults): ED-pool oversubscription for the I/O-bound stages (10 GbE never saturated; est. +10-25% fresh-wall headroom), buffered-read vs mmap for stamp loads (measured ~103 MB/s today), prefetch stage-split and/or depth 3.

Companion issues (same PR): stamp-cache pruning, saved-config allowlist. Also fixes #297 in-PR per maintainer direction.

### LINKED ISSUE #302: [FEATURE]: Automatic stamp-cache pruning with candidate snippet sidecars (default ON)
**Problem.** The #298 fingerprint-scoped stamp cache never evicts. Full-catalog inference writes 5–15 GB of stamps per cadence → **30–90 TB for 6,093 cadences vs 2.7 TB free on /datax** — a naive run fills the disk after ~300–500 cadences. Maintainer-approved direction (HANDOFF follow-up (b)): automatic pruning, default ON.

**Design (maintainer-decided 2026-07-31, "v1 snippet sidecars"):**
- After a cadence's `status='inferred'` manifest row lands, delete its stamp `.npy` (keep the metadata `.json`). Resume rides the DB row — a manifest-skipped cadence never touches its `.npy` (verified by construction); `'failed'` cadences keep stamps until they succeed.
- At scoring time, snapshot each **candidate's** single snippet (~196 KB, `(6,16,512)` float32) into a per-cadence sidecar (atomic tmp → `os.replace`); candidate figures + gallery read sidecar-first. Disk stays bounded at any candidate rate.
- A ~2.4 MB in-memory top-K pixel pool in `InferenceVizCollector` keeps the stamp gallery whole after pruning.
- Knob: `inference.prune_stamps` + `--prune-stamps/--no-prune-stamps`. Default **ON** for the fingerprint-scoped default cache dir; default **OFF** when `--preprocess-output-dir` is explicitly set (operator-curated cache) unless the flag is passed; the resolved mode is logged. Opt-out = "keep everything" for slice-scale runs that want the cross-run rerun cache (the only regime where keeping stamps fits on disk anyway).
- Safety: prune only stamps **this run freshly extracted** (never destroys a cache the run was handed); concurrent same-fingerprint runs race soft-safely (contained failure → retry re-extracts; documented, no cross-process locking).
- **Fingerprint hygiene:** the new knob must join `_INFERENCE_FINGERPRINT_EXCLUDE_INFERENCE_KEYS` — otherwise every existing `'inferred'` row goes stale and every ED cache dir renames (denylist design).

**Trade recorded:** cross-run re-scoring under a new tag re-pays extraction (~5–15 min/cadence) for pruned cadences; same-run resume/retry is unaffected.

Part of the round-2 inference PR (see #301).

### LINKED ISSUE #303: [FEATURE]: Invert saved-config loading to a model-contract allowlist
**Problem.** `cli.apply_saved_config` layers *everything* from a saved training config minus a skip-LIST (3 amendments so far: `checkpoint`, batching/prefetch, `reproducibility`) — backwards: every future host-scoped field is unsafe by default. Confirmed live footgun for the v1 Phase-3 scenario (train on bla0, infer on blpc3): `manager.n_processes` defaults to `cpu_count()` (config.py:43), so the bla0-saved config carries **96** and layers unconditionally onto 32-core blpc3 whether `--config-path` is explicit or HF-resolved — **there is no `--n-processes` CLI flag to rescue it** — 3× oversubscribing the persistent ED pool on exactly the preprocessing-bound stage that dominates the catalog wall. (Note: the top-level path fields do NOT layer today — `to_dict` nests them under `"paths"`, which `apply_saved_config` skips as a non-dataclass miss; the allowlist formalizes that accident into a contract.)

**Design (evaluator-verified, behavior-neutral by construction):** allowlist = (fields the result fingerprints treat as result-affecting) ∪ (model-contract fields):
- `data`: `width_bin`, `downsample_factor`, `time_bins`, `num_observations` (= `_INFERENCE_FINGERPRINT_DATA_KEYS`).
- `inference`: section minus `_INFERENCE_FINGERPRINT_EXCLUDE_INFERENCE_KEYS` minus artifact paths — thresholds/`mc_draws`/`reference_cloud_size` and all ED params keep layering (model-coupled: training certifies the threshold cascade against this model; fingerprints include them so resume stays coherent).
- `beta_vae`: whole section. `rf`: `latent_variant`, `active_dims`, `calibration_active`, `calibration_method` only.
- Never layered: paths, `gpu`, `manager`, `db`, `logger`, `monitor`, `hf`, `training`, `checkpoint`, `reproducibility`, non-geometry `data.*`.
- Startup diff log: every ignored saved field whose value differs from the resolved value logs `saved=X ignored, using Y`.
- Tests: allowlist ⊇ fingerprinted keys (can't drift from run_state.py); `manager.n_processes` never layers (regression).
- Belt-and-braces: add `--n-processes` CLI flag.
- Must land **before** any Phase-3 tag is minted (fingerprint stability); for existing artifacts resolved values are identical, so nothing invalidates.

Part of the round-2 inference PR (see #301).

zachtheyek:
## Summary

Round-2 inference optimization pass, from a 12-agent fresh-eyes audit of the post-#299 pipeline grounded in the real `pipeline_stages` spans of the #298 validation runs (5 domain finders × 5 adversarial verifiers + a design evaluator; 33 findings, 24 confirmed). Ground truth that drove the scope: in the fresh regime the main thread + all 5 GPUs idle **91.5% of wall** on preprocessing; a measured **777 s** of un-spanned startup was the `Database.__init__` `COUNT(*)` storm; in the cached regime **viz was 73% of wall** (114 s of it one figure's cold `/datag` reads); and the viz suite could not survive the 6,093-cadence catalog (OOM-class metadata retention + a figure silently lost past 242 cadences).

Closes #301
Closes #302
Closes #303
Closes #297

## What landed

**#301 — performance (numerics "none" unless stated):**
- `Database.__init__` no longer runs the per-table `COUNT(*)` storm (measured 777 s/cold launch on the 80 GB production DB, re-paid every retry relaunch); startup logs the O(1) size pragma; `get_db_stats` stays on-demand. Schema **v8**: partial index `idx_inference_results_supersede (tag, npy_path) WHERE superseded = 0` — migration-only CREATE (the predicate needs v1's column).
- Viz suite made catalog-survivable: sidecars stream-reduce into bounded `CadenceVizSummary` objects (legacy sidecars' raw lists binned at load); **new sidecar contract** — pre-binned `hit_spectrum_hist` + `bandpass_envelopes` added, the ~19 MB `raw_hit_frequencies_mhz` lists dropped (merged lists stay); preproc funnel capped at 120 bars + aggregate (it exceeded Agg's 65,536-px canvas past 242 cadences and `_viz_safe` swallowed the failure — every catalog-scale pass silently lost it); per-figure `pipeline_stages` sub-spans; `--inference-viz-scope {full,new}` for resumed campaigns; candidate-sidecar reads memoized; monitor teardown plot per-series-decimated (multi-GB RAM spike gone).
- Preprocessing: overlapping/abutting stamp windows fetched as one coalesced wide read (byte-identical, pinned); `_chunk_cache_kwargs` hoisted to the parent (~264 redundant network-FS opens/cadence gone); per-task whole-mapping `msync` removed; sequential already-downsampled loads use one chunk (no ~65 GB concatenate re-copy); NaN check rides the min reduction; ED progress at 25% milestones (the per-channel lines were a measured 62% of Slack-bound log volume; explicit `--coarse-channel-log-interval` restores every-N).
- RF path: `prepare_latent_features` → reshape (float64 kept deliberately — the training-side AU gate `.var()`s this output and float32 moves threshold-deciding bits; equivalence pinned against the verbatim historical loop); viz collector decides the budget before building features (~190 MB/cadence transient gone); `_batched_mc_scores` transient halved; `RandomForestModel.load` re-pins `n_jobs` from the runtime config (the pickle carried the training host's value).
- Encode: one-step software pipeline (dispatch k+1 before harvesting k) + encode-only submodel (the discarded `Sampling` draw no longer executes). **Honest note:** no measured encode-wall change on the cached 2-cadence subset (28.8 s total, unchanged); both changes are numerics-clean and kept for structure, with A/B numbers below.

**#302 — stamp-cache pruning (maintainer-decided design, default ON):** after a cadence's `'inferred'` row + viz collection, its stamp `.npy` is deleted; metadata `.json` always kept; candidate snippets (~196 KB each) snapshotted into atomic `.candidates.npz` sidecars with reader fallback; bounded top-K pixel pool keeps the stamp gallery whole. `None` = AUTO (ON for the fingerprint-scoped default dir, OFF for an explicit `--preprocess-output-dir`); only stamps **this run freshly extracted** are pruned (a resumed run never deletes a handed cache); best-effort containment. Without this, the full catalog writes 30–90 TB vs 2.7 TB free. Trade recorded: cross-run re-scoring re-pays extraction (~5–15 min/cadence) for pruned cadences; same-run resume/retry unaffected.

**#303 — saved-config allowlist:** `apply_saved_config` layers only model-contract/result-affecting fields (derived from `run_state.py`'s fingerprint key sets, drift-test-pinned); everything host/run-scoped is ignored with a startup diff log. Kills the confirmed v1-scenario footgun: a 96-core bla0-saved config layering `manager.n_processes=96` onto 32-core blpc3 with no CLI rescue. New shared `--n-processes` flag (≥ 1 validated). Byte-identical outputs by construction (every de-layered field is in the result-invariant fingerprint denylist).

**#297 —** parse-time floor `num_samples_rf × train_val_split ≥ effective_batch_size` (the run previously died at `rf_train` after all VAE rounds); deliberately `≥`-only — the shipped defaults rely on the runtime trim (79,872 % 7,680 ≠ 0); mirrored into `_check_cross_constraints` so suggested fixes can't propose a dying config; matrix test.

**Fingerprint hygiene (load-bearing):** `prune_stamps` + `inference_viz_scope` joined both `run_state.py` denylists; stability pinned against a master-shaped config dict — no `'inferred'` resume row stales and no ED cache dir renames on upgrade (verified live: the branch resolved the production cache dir byte-for-byte).

## Validation (blpc3, NGC container, weights `train_20260729_125502` exclusively)

- **Unit suite: 1084 passed, 0 failed at the branch tip** (blpc3 container, 3.12) (earlier tip: 1083 passed with one deselection — `test_reg_loss_deterministic_across_calls`, a pre-existing GPU-autotune flake that **fails on master on the same host**; per maintainer direction it is now FIXED in this PR by enabling op determinism around its two calls and restoring the prior process-global state, and passes 3/3 on GPU). CI: see checks.
- **Gate 1 — fresh extraction byte-identity:** subset_test.csv re-extracted through the new code into a scratch dir → both stamp `.npy` files **byte-identical** (md5) to the master-produced production cache; shared metadata keys equal (incl. `ed_config_fingerprint`); new keys present; raw list absent. Tag `inf_20260731_004408`.
- **Gate 2 — fingerprint + prune guard live:** default dir resolved to the exact production cache `subset_test_ed7c3ba8070877`; pruning AUTO=ON yet the handed cache untouched (`freshly_extracted` guard). Tag `inf_20260731_005401`.
- **Det-pair:** vs the recorded master+det run `inf_20260730_100142`: **identical candidate sets (10 = 10), max |Δ| = 0.0** across confidence / screening_proba / mc_mean / mc_std — bit-exact science outputs.
- **Allowlist live:** the real training config's diff log ignored exactly the right fields (`tf_deterministic_ops=False` historic, `num_training_rounds=5`, pre-#298 `per_replica_batch_size=2048`, `checkpoint.save_tag`) while layering the model contract.
- **Viz sub-spans (new instrumentation):** envelope-rendered bandpass figure **0.5 s vs 104.5 s** legacy live-read on the same host — cached-rerun viz drops 141.4 s → ~36 s once sidecars carry envelopes.
- **A/B legs (8-cadence fresh subset, per-leg scratch dirs, pruning exercised live):** three fresh-regime legs on the same 8 /datag cadences as #298's depth A/B, per-leg scratch dirs with pruning ON:
  - **Leg A** (branch defaults, depth 2): **5,118 s** wall vs the recorded master depth-2 run's 5,699 s (~−10% from the round-2 preprocessing work).
  - **Leg B** (`--n-processes 64`): **5,290 s — measured REJECTION** of pool oversubscription (read_ed +36%, extract +44% vs leg A despite a warmer cache; 64 workers thrash 32 cores before the 10 GbE link saturates). `n_processes` stays `cpu_count()`.
  - **Leg C** (`--prefetch-depth 3`): **4,071 s, ~−20% vs leg A** → default flipped to 3 with the honest ~10–20% band documented (leg C ran last/warmest — the same confound structure the 1→2 flip carried).
  - All legs: candidate sets identical to the recorded master run, **0.0 score deltas**; pruning live throughout — each leg's cache dir closed at **61 MB** (metadata + candidate sidecars) instead of ~190 GB of stamps, `/datax` flat at 2.9 T free.
  - Full-catalog projection: ~11.9 → ~8.5–10.7 min/cadence fresh (≈ 7 weeks → ≈ 5.5–6.5 weeks), and the run now survives its own disk/RAM (pruning + the viz rework) — previously it filled the disk after ~300–500 cadences.

## Review notes

- The metadata sidecar contract change (dropped raw hit lists) is deliberate and one-way for NEW sidecars; every reader handles legacy sidecars (reduce-at-load) — no migration needed.
- The hit-spectrum figure is *visually* identical (≥ 40× finer rebinning), not byte-identical; the bandpass figure renders identical lines from stored envelopes (commuting-mean exactness) with float32-despike low-bit caveat on 4 DC bins. All science outputs are bit-exact per the det-pair gate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -924,10 +924,12 @@ options:
                         background (0 disables)
   --prefetch-depth PREFETCH_DEPTH
                         Cadences preprocessed+loaded ahead of the GPU stage in
-                        the streaming loop (>= 1). Depth 2 overlaps one
-                        cadence's energy-detection reads with the previous
-                        one's stamp extraction, costing one extra in-flight
-                        cadence of RAM; outputs are identical at any depth
+                        the streaming loop (>= 1). Each unit of depth overlaps
+                        energy-detection reads with stamp extraction and the
+                        serial per-cadence sections, costing one in-flight
+                        cadence of RAM (up to ~65 GB for RFI-dense C-band
+                        cadences); outputs are identical at any depth
+                        (default: 3 per the on-cluster A/B)
   --cadence-group-by-cols CADENCE_GROUP_BY_COLS [CADENCE_GROUP_BY_COLS ...]
                         Space-separated list of CSV column names whose joint
                         value defines cadence membership (e.g., Target Session
--- a/README.md
+++ b/README.md
@@ -414,9 +414,9 @@ The Aetherscan training pipeline exposes the following CLI flags to the user. Re
 ```
 usage: train [-h] [--seed SEED] [--unseeded]
              [--tf-deterministic-ops | --no-tf-deterministic-ops]
-             [--data-path DATA_PATH] [--model-path MODEL_PATH]
-             [--output-path OUTPUT_PATH] [--dashboard | --no-dashboard]
-             [--dashboard-port DASHBOARD_PORT]
+             [--n-processes N_PROCESSES] [--data-path DATA_PATH]
+             [--model-path MODEL_PATH] [--output-path OUTPUT_PATH]
+             [--dashboard | --no-dashboard] [--dashboard-port DASHBOARD_PORT]
              [--benchmark-report | --no-benchmark-report]
              [--vae-latent-dim VAE_LATENT_DIM]
              [--vae-dense-layer-size VAE_DENSE_LAYER_SIZE]
@@ -498,6 +498,13 @@ options:
                         Default: enabled — without it, cuDNN autotune noise
                         can flip near-threshold candidates between identical
                         runs; opt out with --no-tf-deterministic-ops
+  --n-processes N_PROCESSES
+                        Worker-process count for the multiprocessing pools
+                        (energy detection + stamp extraction at inference;
+                        data generation at training). Default: all cores. Host
+                        tuning: never layered from a saved --config-path, so a
+                        config recorded on a bigger host cannot oversubscribe
+                        this one (must be >= 1)
   --data-path DATA_PATH
                         Path to data directory (overrides AETHERSCAN_DATA_PATH
                         environment variable)
@@ -768,8 +775,9 @@ The Aetherscan inference pipeline exposes the following CLI flags to the user. R
 ```
 usage: inference [-h] [--seed SEED] [--unseeded]
                  [--tf-deterministic-ops | --no-tf-deterministic-ops]
-                 [--data-path DATA_PATH] [--model-path MODEL_PATH]
-                 [--output-path OUTPUT_PATH] [--dashboard | --no-dashboard]
+                 [--n-processes N_PROCESSES] [--data-path DATA_PATH]
+                 [--model-path MODEL_PATH] [--output-path OUTPUT_PATH]
+                 [--dashboard | --no-dashboard]
                  [--dashboard-port DASHBOARD_PORT]
                  [--benchmark-report | --no-benchmark-report]
                  [--num-replicas NUM_REPLICAS]
@@ -801,7 +809,9 @@ usage: inference [-h] [--seed SEED] [--unseeded]
                  [--overlap-search | --no-overlap-search]
                  [--overlap-fraction OVERLAP_FRACTION]
                  [--preprocess-output-dir PREPROCESS_OUTPUT_DIR]
+                 [--prune-stamps | --no-prune-stamps]
                  [--inference-viz | --no-inference-viz]
+                 [--inference-viz-scope {full,new}]
                  [--stamp-gallery-top-k STAMP_GALLERY_TOP_K]
                  [--max-candidate-plots MAX_CANDIDATE_PLOTS]
                  [--max-retries MAX_RETRIES] [--retry-delay RETRY_DELAY]
@@ -828,6 +838,13 @@ options:
                         Default: enabled — without it, cuDNN autotune noise
                         can flip near-threshold candidates between identical
                         runs; opt out with --no-tf-deterministic-ops
+  --n-processes N_PROCESSES
+                        Worker-process count for the multiprocessing pools
+                        (energy detection + stamp extraction at inference;
+                        data generation at training). Default: all cores. Host
+                        tuning: never layered from a saved --config-path, so a
+                        config recorded on a bigger host cannot oversubscribe
+                        this one (must be >= 1)
   --data-path DATA_PATH
                         Path to data directory (overrides AETHERSCAN_DATA_PATH
                         environment variable)
@@ -925,10 +942,12 @@ options:
                         Number of fine channels per coarse channel (default:
                         1048576)
   --coarse-channel-log-interval COARSE_CHANNEL_LOG_INTERVAL
-                        Progress-logging chunk size for energy detection, in
-                        coarse channels per log line (default: the number of
-                        worker processes). Parallelism itself comes from the
-                        persistent worker pool, not this knob.
+                        Progress-logging cadence for energy detection, in
+                        coarse channels per log line. Default: ~25% milestone
+                        lines per ON file (the per-channel lines were 62% of a
+                        run's Slack-bound log volume); pass an explicit N to
+                        restore every-N-channels lines. Parallelism itself
+                        comes from the persistent worker pool, not this knob.
   --bandpass-method BANDPASS_METHOD
                         Bandpass flattening method for energy detection: 'pfb'
                         (default) divides each coarse channel by the
@@ -985,6 +1004,18 @@ options:
                         directory. Pass a directory explicitly to pin/share
                         one location (reuse is still guarded by the sidecar's
                         recorded h5 paths and ED fingerprint)
+  --prune-stamps, --no-prune-stamps
+                        Delete each cadence's stamp .npy right after its
+                        'inferred' manifest row lands, keeping the metadata
+                        .json plus a ~196 KB snippet sidecar per candidate —
+                        resume rides the DB row, and only stamps this run
+                        freshly extracted are ever pruned. Without pruning a
+                        full catalog writes ~30-90 TB of stamps. Default: AUTO
+                        — enabled for the fingerprint-scoped default cache
+                        directory, disabled when --preprocess-output-dir is
+                        set explicitly. Pass --no-prune-stamps to keep every
+                        stamp (slice-scale runs wanting the cross-run rerun
+                        cache).
   --inference-viz, --no-inference-viz
                         Render the inference visualization suite (energy
                         detection distributions, hit spectrum, bandpass
@@ -994,6 +1025,15 @@ options:
                         plots/inference/{save_tag}/ and uploaded to Slack
                         (default: enabled). Pass --no-inference-viz to
                         disable.
+  --inference-viz-scope {full,new}
+                        Which cadences the metadata-driven viz figures cover:
+                        'full' (default) renders the whole accumulated tag
+                        every successful pass; 'new' renders only cadences
+                        inferred this pass — recommended for resumed multi-
+                        pass catalog campaigns, where 'full' re-pays the
+                        entire catalog's viz tail on every pass. DB-sourced
+                        candidate figures always cover the full tag either
+                        way.
   --stamp-gallery-top-k STAMP_GALLERY_TOP_K
                         Number of top-statistic stamps shown in the stamp
                         gallery figure, each as a 6-observation waterfall grid
--- a/docs/CONFIG_AND_CLI.md
+++ b/docs/CONFIG_AND_CLI.md
@@ -37,7 +37,9 @@ parse_args()                       ── argparse picks the specified subparser
    ▼
 apply_saved_config(config_path)    ── only runs on args.command == inference and args.config_path
    │                                  is not None:
-   │                                  overrides the singleton defaults at runtime with a saved JSON
+   │                                  layers the saved JSON's allowlisted model-contract fields
+   │                                  onto the singleton defaults (#303); every other saved field
+   │                                  is ignored (logged when it differs from the resolved value)
    │
    ▼
 validate_args(args)                ── raises ValueError on any failure (no mutation)
@@ -53,11 +55,28 @@ apply_args_to_config(args)         ── writes non-None args onto the Config s
 train_command() / inference_command()
 ```
 
-Priority order:
+Priority order, for the fields a saved config is allowed to layer:
 
 ```
 runtime defaults  <  loaded config  <  CLI args
 ```
+
+`apply_saved_config` is **allowlist-only** (#303): a saved training config layers exactly
+the model-contract / result-affecting fields — the whole `beta_vae` section, the RF's
+deployed-representation contract (`latent_variant` / `active_dims` / `calibration_active` /
+`calibration_method` — not `n_jobs`, not seeds), the data geometry keys, and precisely the
+inference fields the resume fingerprints hash minus the artifact paths. Everything else —
+paths (in nested or legacy flat form), `gpu`, `manager`, `db` / `logger` / `monitor`, `hf`,
+`training`, `checkpoint`, and `reproducibility` — always comes from env + defaults + CLI,
+and every ignored saved field whose value differs from the resolved one is logged as a
+startup diff line, so nothing disappears silently. The old inverse design (apply everything
+minus a skip-list) left every host-scoped field unsafe by default — the live footgun being
+`manager.n_processes`: a config saved on a 96-core host silently 3×-oversubscribed a
+32-core host's worker pool, with no CLI rescue before `--n-processes` existed. The
+allowlist is **derived from `run_state.py`'s fingerprint key sets** and pinned by a drift
+test, so `cli.py` and `run_state.py` can never silently disagree about what is
+result-affecting: a new result-affecting field defaults to layering, a new host field
+defaults to ignored — both fail-safe.
 
 ## The configuration singleton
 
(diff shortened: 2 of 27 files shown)


zachtheyek:
Thanks for the thorough first pass. Two rounds of work since: your six inline notes, then a deeper fresh-eyes audit I ran on the fault-tolerance paths (retry/resume, concurrency, narrow-band catalogs) that the single-pass validation gates structurally couldn't exercise. Container unit suite is green at the branch tip (`fdb6b6f`, blpc3 3.12).

## Your inline notes

**Addressed** (`97c33af`, `7836a91`):
1. **`round(nsr*tvs)` vs the runtime split** — switched to `int()`. You were right this was the closest to a real correctness issue; the runtime floors *per label*, so a fresh-eyes pass then caught that `int(nsr*tvs)` on the total can still exceed the sum of per-label floors by up to `n_labels-1` when `eb % 4 != 0` — noted in the code as the residual, covered by the runtime backstop for exact-floor pathologies (tightening it to the exact per-label sum would risk over-rejecting valid configs).
2. **`_coalesce_stamp_groups` empty-input** — guarded (`return []`).
3. **`lru_cache` aliased dicts** — explicit immutability contract comment at the cache.
4. **`inference_viz_scope` domain check** — added to `collect_validation_errors` (the CLI `choices=` covers the flag path; this covers programmatic/config-file sets), with a test.
5. **`inference.py` encode submodel shape contract** — added a `len(outputs) == 2` check routed into the existing full-encoder fallback, so a future head-count change degrades loudly instead of mis-slicing.

**Rebutted:**
6. **`db.py` decimation pre-flight `COUNT(*) … GROUP BY`** — kept as-is. It runs once at teardown and is strictly cheaper than the full-window materialization it replaces (it reads the same rows the old `_save_plot` did, but returns only the strided subset, so peak RAM drops from all-rows to ≤4,096/series). The `LIMIT/OFFSET` alternative you floated needs a per-series count anyway. If a catalog-scale teardown ever shows it as a stall I'll revisit, but it can't be slower than the code it replaced.

## Fresh-eyes hardening (`b180c63`, `fdb6b6f`)

A 2-run adversarial audit (verified line-by-line) surfaced five real defects on paths no gate covered — no failing cadence, no narrow-band catalog, single pass only:

- **Stamp-cache disk leak** (the important one): a cadence this run extracted, then failed-and-retried, resumed via the artifact path with `freshly_extracted=False`, so pruning skipped it *forever*. Over weeks of transient `/datag`/GPU hiccups that strands multi-GB stamps and refills the disk `#302` exists to keep flat. Fixed: the preprocessor tracks the `npy_path`s *this process* extracted (it outlives the in-process retry loop), so a resumed-own-extraction is prunable while a genuinely handed cache stays exempt.
- **Stamp gallery blanks on retry** (viz-only): the pixel pool was per-attempt, so one transient failure blanked the whole gallery on the retry render. Fixed: a run-scoped pool persists across in-process attempts; cross-*process* relaunch still degrades and is now documented rather than over-claimed.
- **`nan GB` storage** on every default run (funnel + summary): the `.npy` is pruned before the render `getsize`. Fixed: reconstruct the transiently-held size from the stored geometry.
- **Hit-spectrum aliasing** (science figure): new sidecars pre-bin over the full band, so a narrow catalog-hit-span made the 200-bin figure axis finer than the fine grid → a fabricated RFI picket-fence. Fixed: clamp figure bins to the coarsest contributing fine grid (`_clamp_hit_spectrum_bins`, unit-pinned); wide-span common case unchanged.
- **Sidecar tmp leak** (minor): the age-gated sweep globbed only `.tmp.npy`; broadened to catch `.candidates.npz.*.tmp` and the pre-existing `.json.*.tmp`.

Also: added the fingerprint-stability test the body claimed existed (master-shaped == branch-shaped hash, both fingerprints); pinned the `rf`/`beta_vae` allowlist literals (the drift test only derived-and-pinned inference/data); corrected the CONFIG_AND_CLI "a new host field defaults to ignored" sentence (a new `InferenceConfig` field defaults *into* the allowlist + fingerprints and must be denylisted explicitly).

**Deliberately documented, not code-fixed:** concurrent runs sharing the default cache dir with pruning ON can race (delete/overwrite a `.npy`/`.candidates.npz` another is mid-read of) — contained, self-heals via retry. Cross-process file locking was the deferred `#302` design call; the docs now steer concurrent runs to a per-run `--preprocess-output-dir` or `--no-prune-stamps`.

Container unit suite: **1094 passed, 0 failed** at `fdb6b6f` (blpc3, 3.12) — the ten new fault-tolerance/aliasing/allowlist tests included. The subset A/B, det-pair (0.0 score deltas), and byte-identity gates from the PR body are unchanged (these commits touch only the retry/viz/validation surfaces, not the scoring path).

@claude — ready for pass 2.

zachtheyek:
Thanks — pass 2 LGTM noted, and the one residual note is fixed in `1e6f6ba`.

**Encode head-count guard (`inference.py:282`)** — you were exactly right that the check was on the wrong side of the slice (the derived model is `encoder.outputs[:2]`, so its count is always ≤ 2 and the 3→4 add-a-head case slipped through). One correction on the suggested literal: the VAE encoder is `keras.Model(inputs, [z_mean, z_log_var, z])` — **3 heads** (`vae.py:559`), not 2, so `!= 2` would fall back on every real encoder and disable the E6 sampling-skip entirely. The guard now checks the source encoder's count against **3** (we drop the third, sampled-`z`, head). I also extracted the logic into `_derive_encode_model` and added the fallback-path test that was missing — a 2- or 4-head fake encoder falls back, a 3-head one yields the 2-output submodel. `test_inference.py`: 27 passed on blpc3 (container).

That closes every note across both passes. @claude — final look?

Issue #10, opened 2026-01-02
Deprecate TensorBoard support until further notice


zachtheyek:
## Summary

TensorBoard support is being deprecated from the Aetherscan training pipeline until further notice.

## Motivation

The TensorBoard integration added complexity to the training pipeline without providing sufficient value for the current development workflow. The training history is already being tracked in the `self.history` dictionary and visualized through matplotlib plots at checkpoints.

## Changes

All TensorBoard-related functionality has been commented out rather than deleted to facilitate potential future revival if needed:

- **`_setup_tensorboard_logging()` method**: Commented out the entire method that created TensorBoard writers and log directories
- **`__del__()` method**: Commented out the destructor that cleaned up TensorBoard writers
- **`train_round()` logging**: Commented out the per-epoch metric logging to TensorBoard (11 scalar metrics)
- **`archive_directory()` function**: Renamed from `handle_directory()` with comments explaining the incomplete TensorBoard directory handling functionality

## Affected Components

- `src/aetherscan/train.py` - All TensorBoard integration code
- Training history is still maintained in `self.history` dictionary
- Matplotlib plots are still generated at checkpoints via `plot_beta_vae_training_progress()`

## Future Work

If TensorBoard support needs to be revived:
1. Uncomment all code marked with `# COMMENTED OUT: Removing TensorBoard support`
2. Update `archive_directory()` to properly handle TensorBoard log directories (set `target_dirs=["train", "validation"]`)
3. Test the integration thoroughly

## Related Files

- `src/aetherscan/train.py`

zachtheyek:
Some more color: the interplay between TensorBoard logs & fault tolerance was too much to handle considering the value it provided was minimal, especially with the addition of the SQLite db & more frequent checkpoint plots. TensorBoard connection also had to be maintained through SSH tunneling, which was unreliable & finicky. Finally, TensorBoard also lacks many "modern" features that I'd expect from a real-time training/inference monitoring setup. Will revisit in the future if there's enough demand, but for the time being, it's a smarter decision to deprecate support entirely.

PR #309, opened 2026-08-01
perf: round-3 audit — inference float32 features, producer/DB/viz efficiency, cadence freq-width validation

### LINKED ISSUE #308: perf: round-3 fresh-eyes audit — inference float32 features, producer/DB/viz memory + throughput, cadence freq-width validation
### Summary

A fresh-eyes performance & correctness audit of the train/inference pipelines (post #299/#305)
surfaced a set of **net-new, numerics-neutral** optimizations plus one robustness hardening. This
umbrella issue tracks the ones approved for implementation; each was adversarially verified to be
byte-identical (no change to any detection outcome) before landing. The core science/logic is
untouched — these only make the pipeline faster and leaner.

### Approved items (this issue)

- **A — inference builds RF features in float32.** `prepare_latent_features` returned float64
  unconditionally; inference then cast everything back to float32 (and sklearn casts anyway), so
  two full-matrix float64 widenings per cadence were pure waste on the GPU-idle RF stage. Add a
  `dtype` param (default float64 — training's #288 Active-Units `.var()` gate still needs it) and
  pass float32 at inference. Byte-identical downstream.
- **B — free non-winner variant forests after selection (training).** The 7 non-winner
  1000-tree forests stayed resident through the calibration/MC/SHAP/metrics tail; they are already
  persisted to disk and never re-read. Drop them right after the winner is chosen. Memory-only.
- **C — memoize Stage-A background intensity stats (training data-gen).** Stage A is a pure
  function of the raw background but was recomputed ~33×/round per background. Memoize per
  `background_index` on the worker path only (guarded on the immutable shared block). Plus two
  micro-allocs (reshape-stack instead of `zeros`+loop; `copy=False` on the stats-only float64
  cast). Byte-identical `injection_stats` rows.
- **E — validate frequency-width consistency across the 6 cadence files (preprocessing).**
  The geometry check validated rank + time bins but not frequency width, while extraction reads
  the same primary-derived channel window from all 6 files. A narrower file crashed cryptically
  deep in a worker; add `obs_shape[-1]` to the skip-and-warn check. No-op for valid cadences.
- **H — retain bandpass envelopes on one cadence only (inference viz).** Only the first cadence's
  envelopes are ever plotted, but every cadence's were held for the whole render (~GB at catalog
  scale). Keep the first, null the rest. Byte-identical figure.
- **G — `PRAGMA temp_store=MEMORY` (DB).** Keeps teardown/plot GROUP BY/ORDER BY/window sorts in
  RAM instead of spilling to disk. (`mmap_size` deliberately skipped — SIGBUS risk on shared
  nodes + marginal under the per-query connection model.)

### Deferred to a follow-up (need benchmarking / carry more risk)

- Cadence-level encode↔RF/DB software pipeline (reproducibility-sensitive; prototype first).
- Parallelize the viz metadata-sidecar reduction (helps only very large catalogs).
- RF-encode concat-only gather (delicate index alignment; modest one-time win).

### Explicitly not doing

- bf16 mixed precision — confirmed a science change (moves detection bits), correctly stays off.
- Dropping the constant `injection_stats.metadata` column — maintainer chose to keep per-row
  provenance.

### Methodology

Fresh-eyes finder pass (7 domain agents, ledger-grounded to net-new items) → maintainer triage →
adversarial verification (one skeptic per finding, all confirmed byte-identical). Ships with unit
tests for each behavioural surface.

### LINKED ISSUE #310: perf follow-up (#308): RF-encode concat-only gather (#10 implemented); #2 + #8 evaluated-and-declined
### Summary

Follow-up to #308 (the round-3 fresh-eyes audit), which deferred three larger items pending
benchmarking. With the Phase-3 subset run's measured stage profile now in hand, the disposition is:

- **#10 — RF-encode gathers only the concat block: IMPLEMENTED** (in the PR linked below). The RF
  feature encode reused the 3-way training dataset (concat + true + false gathered and copied to
  device every step) but its `encode_fn` consumes only concat. A `concat_only` flag on
  `prepare_distributed_train_dataset` gathers concat alone, cutting the one-time RF-encode
  gather + host→device transfer ~3× → 1× (~59 GB → ~19.6 GB at production shapes). Index
  generators and the concat gather are untouched → byte-identical batch order and
  `train_indices`/`val_indices` alignment.

- **#2 — cadence-level encode↔RF/DB software pipeline: DECLINED (removed from backlog).** The
  measured run is heavily preprocessing-bound (encode + rf + db_write ≈ 2,088 s of mostly-already-
  overlapped work vs ~49,600 s of preprocessing across the 32-worker pool + prefetch). Overlapping
  encode(i+1) with rf/db(i) therefore buys ~nothing in the production/fresh regime, and its
  background RF thread (`n_jobs=-1`) would contend with the preprocessing pool — plausibly
  net-negative — while carrying reproducibility risk. Not worth it.

- **#8 — parallelize the viz metadata-sidecar reduction: DECLINED.** Measured
  `inference.viz.load_metadata` = **8.37 s for 350 cadences** → ~2–3 minutes even at the full
  6,093-cadence catalog, a rounding error on a preprocessing-bound run that takes weeks. Doing it
  correctly also requires splitting the reduction into a new TF-free module (mirroring
  `candidate_figures`) because `inference_viz` pulls in TF+setigen transitively — disproportionate
  surface area for the payoff.

### Disposition

#10 lands in the linked PR; #2 and #8 are recorded here as evaluated-and-declined so they don't
linger as backlog. Reopen only if a cached-rerun-dominated or finalize-bound workload ever makes
#2/#8 material.

zachtheyek:
Closes #308

## What & why

A third fresh-eyes performance/correctness audit of the pipeline (after #299 and #305) surfaced a
set of **net-new, numerics-neutral** optimizations plus one robustness hardening. Every change was
adversarially verified to be **byte-identical** — no detection outcome moves. The core
science/logic is untouched; this is purely faster/leaner.

Methodology: 7 ledger-grounded finder agents → maintainer triage of the trade-offs → one
adversarial skeptic per surviving finding (all confirmed `numerics=none`). Each behavioural
surface ships a unit test.

## Commits (one per finding, all GPG-signed)

| Commit | Finding | Effect | Numerics |
|---|---|---|---|
| `858adb6` | **A** inference float32 RF features | skips two per-cadence float64 widenings on the GPU-idle RF stage | none (float32→float64→float32 round-trips exact; training default stays float64 for the #288 AU-gate) |
| `ca8a7c5` | **B** free non-winner variant forests | training peak RSS through the calibration/MC/SHAP tail | none (memory-only) |
| `860440a` | **C** memoize Stage-A background stats + gen micro-allocs | producer throughput (Stage A recomputed ~33×/round → once) | none (byte-identical `injection_stats`) |
| `396fb18` | **E** freq-width geometry validation | cryptic deep-worker crash → clean skip-and-warn | none (no-op for valid equal-width cadences) |
| `19364da` | **H** one-cadence bandpass-envelope retention | ~GB viz RAM at catalog scale | none (byte-identical figure) |
| `c861a7f` | **G** `temp_store=MEMORY` | teardown/plot sorts stay in RAM | none |

### Safety notes worth reviewing

- **A**: verified inference never runs the `.var()` Active-Units gate (`active_dims` comes from the
  saved config); the float64 default is preserved for training and pinned by
  `test_reshape_matches_historical_loop_and_keeps_float64`.
- **C**: the memo is consulted **only** on the worker path where the plate IS the immutable
  `_GLOBAL_BACKGROUNDS` block (identity-guarded), cleared on each fresh attach, and returns a fresh
  dict per call — the in-process/test path is never cached, so no foreign plate's stats can leak.
  The stacking `astype` stays a copy (setigen mutates it in place); `copy=False` is only on the
  read-only stats cast.
- **E**: rejects only genuinely mismatched cadences; equal-width (the real-data norm) is a no-op.
- **G**: `mmap_size` deliberately **not** set (SIGBUS risk on shared cluster nodes + benefit lost
  to the per-query open/close model).

## Deferred (follow-up)

Cadence-level encode↔RF/DB pipeline (reproducibility-sensitive), parallel viz metadata reduction
(large-catalog only), RF-encode concat-only gather (delicate) — all need benchmarking, tracked for
a separate PR. bf16 confirmed staying off (science change); dropping the `injection_stats.metadata`
column was declined by the maintainer (keep per-row provenance).

## Validation

- Full container unit suite: **1106 passed** on the NGC image (bla0), `not gpu and not cluster`.
- Ruff lint + format clean.
- Adversarial-verify pass: all 6 findings `numerics=none`, byte-identical.
- (Context) the v1 Phase-3 inference run is proceeding on blpc3 against `train_20260729_152426`
  on current master — unaffected by this branch.

---

## Update — follow-up items (Closes #310)

Per maintainer decision, the three deferred items (#308) were dispositioned with the completed run's measured stage profile:

- **#10 — RF-encode concat-only gather: ADDED to this PR** (`85e34ef`). `prepare_distributed_train_dataset` gains a `concat_only` flag so the RF feature encode gathers only the concat block (it discards true/false) — cutting that one-time gather + host→device transfer ~3× → 1× (~59 GB → ~19.6 GB). Index generators + concat gather untouched → batch order and `train_indices`/`val_indices` alignment byte-identical (pinned by `test_concat_only_yields_concat_block_in_identical_order`).
- **#2 — encode↔RF/DB overlap: DECLINED (removed from backlog).** Measured run is preprocessing-bound (~49,600 s preprocessing vs ~2,088 s mostly-already-overlapped encode+rf+db); the overlap buys ~nothing and its background RF thread would contend with the 32-worker preprocess pool — net-negative, plus reproducibility risk.
- **#8 — parallel viz metadata reduction: DECLINED.** `inference.viz.load_metadata` = 8.37 s / 350 cadences → ~2–3 min at full-catalog scale, negligible on a weeks-long preprocessing-bound run; and it would need a new TF-free module split. Disproportionate.

Validation: full container unit suite green on the branch tip (incl. the new #10 test); the 350-cadence subset re-run on the branch reproduces the master run's candidate set (numerics-neutral gate).

### DIFF
--- a/src/aetherscan/data_generation.py
+++ b/src/aetherscan/data_generation.py
@@ -45,6 +45,12 @@ _GLOBAL_BACKGROUNDS = None
 _GLOBAL_SHAPE = None
 _GLOBAL_DTYPE = None
 
+# Per-process memo of Stage-A (raw-background) intensity stats, keyed by background_index.
+# Only consulted on the worker path, where the plate handed to create_* IS this immutable
+# shared-memory block — see _stage_a_stats. Cleared whenever a worker (re)attaches its
+# backgrounds so a recycled process can never serve a previous block's stats.
+_STAGE_A_STATS_CACHE: dict[int, dict[str, float]] = {}
+
 
 def _init_worker(shm_name, shape, dtype, log_queue=None):
     """
@@ -76,6 +82,9 @@ def _init_worker(shm_name, shape, dtype, log_queue=None):
     _GLOBAL_BACKGROUNDS = np.ndarray(shape, dtype=dtype, buffer=_GLOBAL_SHM.buf)
     _GLOBAL_SHAPE = shape
     _GLOBAL_DTYPE = dtype
+
+    # Fresh backgrounds -> drop any Stage-A stats memoized against a prior attachment.
+    _STAGE_A_STATS_CACHE.clear()
 
     # Ignore SIGINT (Ctrl+C) in workers - let manager from parent handle cleanup coordination
     signal.signal(signal.SIGINT, signal.SIG_IGN)
@@ -365,8 +374,11 @@ def _compute_intensity_stats(data: np.ndarray) -> dict[str, float]:
             float("nan"),
         )
 
-    # Temporarily promote to float64 to prevent overflow in higher-order moments (especially for pre-normalization stats)
-    flat = data.ravel().astype(np.float64)
+    # Temporarily promote to float64 to prevent overflow in higher-order moments (especially for
+    # pre-normalization stats). copy=False avoids a redundant copy when `data` is already float64
+    # (Stage B/C inputs) — `flat` is only ever read here (median/mean/std/abs/skew/kurtosis all
+    # produce new arrays), so aliasing `data` is safe.
+    flat = data.ravel().astype(np.float64, copy=False)
     median_val = np.median(flat)
 
     stats = {
@@ -379,6 +391,27 @@ def _compute_intensity_stats(data: np.ndarray) -> dict[str, float]:
     }
 
     return stats
+
+
+def _stage_a_stats(plate: np.ndarray, background_index: int, base: np.ndarray) -> dict[str, float]:
+    """Stage-A (raw-background) intensity stats, memoized per background_index on the worker path.
+
+    Stage A is a pure function of the raw background, but each background is drawn ~n_samples /
+    n_backgrounds times per round (~33x at production defaults), so recomputing its two median
+    sorts + skew/kurtosis every draw is pure redundancy. In a pool worker the `plate` handed to
+    create_* IS `_GLOBAL_BACKGROUNDS` — a single block loaded once and never mutated for the
+    process's life — so caching by index there is byte-identical and safe. Every other caller
+    (the in-process RF-dataset path, tests) passes a different array and is NOT cached, so no
+    stale/foreign plate's stats can ever leak. A fresh dict is returned each call (callers store
+    or .copy() it) so the cached entry is never mutated in place.
+    """
+    if _GLOBAL_BACKGROUNDS is not None and plate is _GLOBAL_BACKGROUNDS:
+        cached = _STAGE_A_STATS_CACHE.get(background_index)
+        if cached is None:
+            cached = _compute_intensity_stats(base)
+            _STAGE_A_STATS_CACHE[background_index] = cached
+        return dict(cached)
+    return _compute_intensity_stats(base)
 
 
 def create_false(
@@ -409,8 +442,9 @@ def create_false(
     n_time = plate.shape[2]
     final = np.zeros((n_obs, n_time, width_bin))
 
-    # STAGE A: Pre-injection, pre-normalization (raw background)
-    stats_a = _compute_intensity_stats(base)
+    # STAGE A: Pre-injection, pre-normalization (raw background). Memoized per background on
+    # the worker path (see _stage_a_stats) — same background, same stats, computed once.
+    stats_a = _stage_a_stats(plate, background_index, base)
 
     # Initialize signal info (will be populated if inject=True)
     signal_info = {}
@@ -423,11 +457,10 @@ def create_false(
     slope_was_clamped = False
     if inject:
         # Prepare data for signal injection by stacking all 6 observations vertically
-        # (6, 16, 512) -> (96, 512)
-        # Obs 0: rows 0-15, Obs 1: rows 16-31, Obs 2: rows 32-47, ...
-        data = np.zeros((n_obs * n_time, width_bin))
-        for i in range(n_obs):
-            data[i * n_time : (i + 1) * n_time, :] = base[i, :, :]
+        # (6, 16, 512) -> (96, 512). base is C-contiguous, so the row-major reshape IS that
+        # obs-major stack (row i*n_time+t = base[i, t]); .astype(float64) yields the fresh,
+        # independent float64 buffer setigen injection then mutates in place (must be a copy).
+        data = base.reshape(n_obs * n_time, width_bin).astype(np.float64)
 
         # Select a random SNR from the given range & inject RFI into all 6 observations
         snr = random.random() * snr_range + snr_base
@@ -495,15 +528,15 @@ def create_true_single(
     n_time = plate.shape[2]
     final = np.zeros((n_obs, n_time, width_bin))
 
-    # STAGE A: Pre-injection, pre-normalization (raw background)
-    stats_a = _compute_intensity_stats(base)
+    # STAGE A: Pre-injection, pre-normalization (raw background). Memoized per background on
+    # the worker path (see _stage_a_stats) — same background, same stats, computed once.
+    stats_a = _stage_a_stats(plate, background_index, base)
 
     # Prepare data for signal injection by stacking all 6 observations vertically
-    # (6, 16, 512) -> (96, 512)
-    # Obs 0: rows 0-15, Obs 1: rows 16-31, Obs 2: rows 32-47, ...
-    data = np.zeros((n_obs * n_time, width_bin))
-    for i in range(n_obs):
-        data[i * n_time : (i + 1) * n_time, :] = base[i, :, :]
+    # (6, 16, 512) -> (96, 512). base is C-contiguous, so the row-major reshape IS that
+    # obs-major stack (row i*n_time+t = base[i, t]); .astype(float64) yields the fresh,
+    # independent float64 buffer setigen injection then mutates in place (must be a copy).
+    data = base.reshape(n_obs * n_time, width_bin).astype(np.float64)
 
     # Select a random SNR from the given range & inject ETI
     snr = random.random() * snr_range + snr_base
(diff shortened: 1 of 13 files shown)


zachtheyek:
Thanks for the thorough pass — appreciate the independent verification of all six safety claims.

**On the two nits (both non-blocking):**
- `_stage_a_stats` returning `dict(cached)`: intentional and load-bearing exactly as you note — kept and documented, no change.
- The `test_dtype_float32_...` reshape-view point: fair. I'd already strengthened that test in `bba6956` (pushed after your review) to assert byte-identity **through a real consumer** — `build_variant_features("z_mean_logvar", ...)` built from the float32 prepare-outputs vs the float64 ones must be `array_equal`. That pins the property the perf claim actually depends on (every consumer casts to float32), not just the reshape round-trip.

**Heads-up: I ran a fresh-eyes self-audit of the diff before you landed, and pushed `bba6956` with five refinements** (all verified, no blocker/major). Most are cosmetic, but one is a deliberate behavioural refinement to **Finding E** that diverges from the reject-both version you reviewed — flagging it explicitly:

- **E is now `< n_chans`, not strict width equality.** I moved `n_chans` (the `[0:n_chans]` window every file is actually read through) above the loop and reject only files with **fewer** than `n_chans` channels. Rationale for allowing wider files rather than rejecting them: a file with `>= n_chans` channels reads that window cleanly and **processed correctly on master**, so rejecting it would change a currently-successful outcome — against this PR's science-untouched mandate. It's also consistent with how the *primary* is already treated: when a header's `nchans < data_shape[-1]`, the primary itself is "wider than the read window" and the pipeline reads `[0:n_chans]` from it without complaint, so OFF files shouldn't be held to a stricter rule. Net: E now converts only the genuine narrow-file crash into a clean skip, and touches no outcome that works today. (A width check can't detect a differing grid *origin* either way — that's a pre-existing, out-of-scope concern.) Tests split into narrow-skip (parametrized) + a wider-allowed positive test.
- Also in `bba6956`: `dtype: npt.DTypeLike` via an explicit `import numpy.typing as npt` (`np.typing` isn't auto-loaded on `numpy<2.0`, so the bare attribute was an unresolvable name under runtime annotation introspection); `del clf` after the winner-forest reduction (the fit-loop variable still bound the *last* variant's forest, so the tail was carrying 2, not 1); and unit coverage for `_init_worker`'s cache-clear (finding C's sole contamination guard) and the H render path.

Full suite green on the new tip (`bba6956`): **1108 passed** in the NGC container. Re-tagging for a second pass on the refined diff.

@claude

zachtheyek:
Thanks — and thanks for independently confirming the E `< n_chans` rationale, that's exactly the reasoning.

Addressed all three optional nits in `08c1c47`:
- **Comment reflow (`preprocessing.py`)**: fixed the mid-sentence "skipped rather than" break and collapsed the redundant `n_chans` restatement into a single trailing clause.
- **Test teardown hygiene**: `test_init_worker_clears_stage_a_cache` now nulls `data_generation._GLOBAL_BACKGROUNDS` and `_GLOBAL_SHM` in the `finally` (after `shm.unlink()`), so no later test can index a view over freed shared memory. Good catch — safe today but genuinely fragile.

Affected test files green in the container (`189 passed`); full suite already at `1108 passed`. That's all three — purely cosmetic + test-suite hygiene, no source-behavior change.

@claude

zachtheyek:
Re-requesting review — the PR was extended since the last (clean) pass, per maintainer direction on the #308 deferred items.

**New since pass 3:**
- **#10 added** (`85e34ef`): `concat_only` flag on `prepare_distributed_train_dataset` so the RF-feature encode gathers only the concat block (skipping the true/false gather + host→device transfer, ~59 GB → ~19.6 GB one-time). Index generators + concat gather untouched → batch order and `train_indices`/`val_indices` alignment are byte-identical; the RF `encode_fn` destructures the single-tensor element. Pinned by `test_concat_only_yields_concat_block_in_identical_order`.
- **#2 removed / #8 declined** (Closes #310): both evaluated against this run's measured stage profile — #2 (encode↔RF overlap) buys ~nothing in the preprocessing-bound regime and would contend with the preprocess pool; #8 (parallel viz metadata reduce) targets an 8.37 s/350-cad finalize step (~2–3 min at full catalog, negligible on a weeks-long run) and would need a TF-free module split. Recorded in #310.
- **Self-audit fix** (`5c9e693`): my fresh-eyes pass on the #10 diff caught that the new test built the full vs concat_only datasets with `rng=None` → different stratified splits → the assertion failed (validating nothing). Fixed by seeding both builds with a fresh `default_rng(0)`; production #10 code confirmed byte-identical.

CI green on `5c9e693` (tests 3.10/3.11/3.12 + analyze + codeql); full container suite **1109 passed**. Branch smokes + the 350-subset numerics-neutral re-run against `train_20260729_152426` are in flight on blpc3.

Please review the incremental #10 change (esp. the concat_only alignment invariant) and the #2/#8 dispositions.

@claude

zachtheyek:
Final validation complete — merging on the maintainer's behalf (explicitly authorized to merge once review loops pass + no blockers remain):

- **Numerics gate: byte-identical.** 350-cadence subset re-run on the branch (`inf_20260731_225205`) vs the master-code run (`inf_20260731_182011`): identical candidate sets (569 candidates / 48 cadences each; 0 rows only-in-branch, 0 only-in-master) and **max |confidence Δ| = 0.0** across all 569.
- Review loop clean (pass-4 LGTM, no notes); CI green on `5c9e693`; container unit suite 1109 passed; branch smokes passed against the v1 weights.
- Closes #308 #310 (#2/#8 evaluated-and-declined with measured rationale; #10 implemented).

PR #313, opened 2026-08-01
release: v1.0.0

### LINKED ISSUE #183: Update README System Requirements with real numbers from production benchmarking
The **System Requirements** section in `README.md` (lines ~30-58) carries placeholder numbers that are not grounded in real benchmarking — the README itself flags this with an inline `# TODO: update system requirements after proper benchmarking` (README.md:58). Update them from actual runs.

## Currently-inaccurate figures
- `≥12 GB combined VRAM (training) / ≥9 GB combined VRAM (inference)` (README.md:40)
- `≥150 GB RAM (training) / ≥100 GB RAM (inference)` (README.md:41)

These are rough guesses. Both the container and conda blocks reference them (the conda block says "VRAM / RAM same as above"), so both need updating.

## Where the real numbers come from
The pipeline now has always-on resource + stage instrumentation (benchmarking suite, #134) that makes this measurable:
- **RAM**: `system_resources` DB rows (`resource_type='ram'`) captured at 1 Hz during a run; `utils/benchmark_report.py` and the live dashboard plot peak/steady RAM. The §5B bla0 e2e smoke saw ~23 GB main RSS but at tiny (smoke) geometry — **production geometry will differ**, so pull the numbers from a full-scale run.
- **VRAM**: `system_resources` `gpu` rows (per-GPU utilization/memory) + `nvidia-smi` during a real distributed run; depends on model geometry, `global_batch_size`, and replica count.
- **Disk / scratch**: not currently stated but should be — training round data is ~294 GB on `/datax` (see `round_data.py`), plus checkpoints/outputs; inference needs the model trio + cadence data. Worth a "storage" line.

## Suggested action
Derive the real figures from the **final full-scale v1 training + inference runs** (the release-qualification runs — see `docs/RELEASE.md`), i.e. read the `system_resources` data / `benchmark_report` for those runs, then:
- Update the VRAM / RAM figures in README (both the container and conda blocks) to real, measured values (ideally note the config they were measured under — replicas, batch size).
- Add a storage/scratch requirement line.
- Reconcile with `docs/GPU_RUNTIME_GUIDE.md` if it restates any of these.
- Remove the inline `# TODO` at README.md:58 once done.

**Timing:** best done right after the final release training run, when real production numbers exist. (Data-gen throughput context already benchmarked: `data_gen_task_size=64` confirmed optimal at production scale on both clusters — but that's throughput, not the RAM/VRAM the sysreqs need.)

zachtheyek:
Closes #183

## v1.0.0 release PR (docs/RELEASE.md step 4)

Cuts the first stable release. Version is single-sourced in `pyproject.toml`; the CD workflow enforces that the pushed `v1.0.0` git tag matches it exactly.

### Changes
- **`pyproject.toml`**: `version` `0.9.0.dev0` → **`1.0.0`**; Development Status `2 - Pre-Alpha` → **`5 - Production/Stable`** (per the maintainer — v1.0.0 ships blessed release weights + a stable CLI/config contract).
- **README System Requirements (#183)**: replaced the placeholder VRAM/RAM guesses with numbers **measured from the v1.0.0 release runs** via the always-on `system_resources` instrumentation, and dropped the stale "update after benchmarking" TODO:
  - VRAM ~6 GB/GPU (training) / ~2.5 GB/GPU (inference) → **≥8 GB/GPU** recommended
  - RAM ~260 GB training peak (mean ~150) / ~200 GB inference peak at default `--prefetch-depth 3` (mean ~36; tunable) → **≥256 GB** for full-scale
  - Disk ~150 GB/retained round (up to ~3 TB with `--keep-round-data`); inference stamps auto-pruned by default
  - Provenance: training `train_20260729_152426` (6× RTX A4000 16 GB); inference `inf_20260731_182011` (5× RTX PRO 6000 96 GB, 350-cadence `/datag` subset)
- **`CITATION.cff`**: `version` `0.1.0` → `1.0.0`, `date-released` → `2026-08-01`.

### Release sequence (this PR = step 4)
After this merges: TestPyPI dry-run (`workflow_dispatch test_pypi:true`) → **bless** the HF weights (`hf_tag_release.py --save-tag train_20260729_152426 --release v1.0.0`) → **signed tag** (`git tag -s v1.0.0 && git push origin v1.0.0`, maintainer) → CD (guard → tests → HF-verify → build → PyPI → GitHub Release) → `pip install aetherscan==1.0.0` smoke.

No source-code changes; `__version__` continues to read the single-sourced version via `importlib.metadata`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,9 +18,9 @@ name: Release
 on:
   push:
     tags: ["v*"]
-  # Dry-run path (recommended before the first real release): manual dispatch publishes to
-  # test.pypi.org. Real releases only ever happen via tag pushes — a manual run with
-  # test_pypi unchecked fails the guard.
+  # Dry-run path (recommended before EVERY release — docs/RELEASE.md runbook step 4): manual
+  # dispatch publishes to test.pypi.org. Real releases only ever happen via tag pushes — a manual
+  # run with test_pypi unchecked fails the guard.
   workflow_dispatch:
     inputs:
       test_pypi:
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -14,8 +14,8 @@ authors:
     given-names: "Matt"
     orcid: "https://orcid.org/0000-0002-7042-7566"
 title: "Aetherscan: Breakthrough Listen's first end-to-end production-grade DL pipeline for SETI @ scale"
-version: 0.1.0
-date-released: 2026-01-01
+version: 1.0.0
+date-released: 2026-08-01
 url: "https://github.com/zachtheyek/Aetherscan"
 repository-code: "https://github.com/zachtheyek/Aetherscan"
 keywords:
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,4 +1,5 @@
 cff-version: 1.2.0
+type: software
 message: "If you use this software, please cite it as below."
 authors:
   - family-names: "Yek"
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Aetherscan supports two install paths off the same source tree. The NGC containe
   - Blackwell (sm_120, e.g. RTX PRO 6000) — driver ≥570 (native CUDA 12.8)
   - Ampere (sm_86, e.g. RTX A4000) — driver ≥550 (host CUDA 12.3) via CUDA forward compatibility
 - VRAM: **≥8 GB per GPU** recommended — measured peaks ~6 GB/GPU (training) and ~2.5 GB/GPU (inference) on the v1.0.0 release runs; gradient accumulation keeps per-GPU VRAM low
-- RAM: **≥256 GB** for full-scale training and default catalog-scale inference — measured peaks ~260 GB (training; mean ~150 GB) and ~200 GB (inference at the default `--prefetch-depth 3`; mean ~36 GB). Inference RAM scales with `--prefetch-depth` × the largest in-flight cadence, so lower `--prefetch-depth` for smaller-RAM hosts or small catalogs
+- RAM: **≥288 GB** for full-scale training and default catalog-scale inference (measured peaks ~260 GB training / ~200 GB inference, plus headroom — a strict-256 GB host sits too close to the training peak and risks OOM under page-cache pressure). Means are much lower (~150 GB training / ~36 GB inference); inference RAM scales with `--prefetch-depth` × the largest in-flight cadence, so lower `--prefetch-depth` for smaller-RAM hosts or small catalogs
 - Disk: full-scale training round data ~150 GB per retained round (float16 default), up to ~3 TB with `--keep-round-data`; inference stamps are auto-pruned by default (~1 MB/cadence metadata retained + a transient ~5–20 GB/cadence × `--prefetch-depth` during extraction)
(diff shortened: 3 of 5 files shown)


zachtheyek:
Thanks — addressed all three in `b09c7a3`:

- **RAM headroom (finding 1):** good catch — recommending ≥256 GB while the measured training peak is ~260 GB was self-contradictory and unsafe (psutil "used" excludes reclaimable cache, so 260 is the real competing figure). Bumped to **≥288 GB**, which clears both peaks (260 train / 200 inf) with headroom, and I kept the means (~150 / ~36 GB) + the `--prefetch-depth` tuning note.
- **CITATION `type: software` (finding 2):** added.
- **Post-release dev-version reset (finding 3):** rather than leave it implicit, added **step 8** to the `docs/RELEASE.md` runbook — after a release lands, a follow-up chore PR bumps `master`'s `version` to the next `.devN` so it stops advertising as a shipped stable between releases. (Correctly out of scope to do *in* this PR — `version` must stay exactly `1.0.0` for the tag/version guard.)

Nits acknowledged: the 2→5 classifier leap is deliberate (maintainer call, justified inline); `date-released` matches today and I'll adjust if the tag is cut on a different day; the conda block's "same as above" stays correct.

CI green on `b09c7a3`. Re-requesting a confirmation pass.

@claude

zachtheyek:
Added one more commit (`1d5e2b9`) at the maintainer's request: rewrite the `docs/RELEASE.md` runbook to match the **actual** first-release process rather than the original best-guess, and make it idiot-proof.

Key changes:
- Reordered to the real sequence (9 steps): prereqs → train → release PR → **TestPyPI dry-run** → bless → signed tag → CD → smoke → dev-version reset — each with role labels (maintainer / agent / CD) and exact copy-pasteable commands.
- Documented the dry-run in full: `gh workflow run release.yml -f test_pypi=true --ref master` **after** the release PR merges (so it builds the real X.Y.Z), and that it **skips** the signed-tag gate + HF-verify + GitHub Release → needs no bless and no tag (verified against `release.yml`'s `if: dry_run != 'true'` conditions).
- Expanded one-time setup (TestPyPI trusted publisher + `testpypi` environment + how to verify), and captured the gotchas we actually hit: merge-commit-only + `BLOCKED`-until-approval (admin-merge escape hatch), version-single-source == pushed tag, HF-verify-fail recovery (bless then rerun, don't retag), and that the #183 numbers come from the `system_resources` rows.
- **Fixed a real doc bug** while there: the post-install command is `python -m aetherscan.main inference`, not `aetherscan.main inference` — there's no such console script (only `aetherscan-dashboard` exists in `[project.scripts]`), so the old TL;DR + smoke command wouldn't run. Corrected both, and renumbered the step-3→step-5 cross-references for the bless.

Doc-only; no code touched. Please review the runbook for accuracy/completeness.

@claude

zachtheyek:
Addressed both non-blocking nits in `459e99f`:

- **Step 6 tag-target hedge:** added the belt-and-suspenders instruction — if unrelated work merges to `master` after the release PR, tag the release PR's merge commit explicitly (`git tag -s vX.Y.Z <sha>`) rather than `HEAD`, so the release captures exactly the reviewed tree. Squarely in the idiot-proof spirit.
- **`release.yml` dry-run comment:** was "recommended before the first real release" (implied one-time); the runbook promotes it to every release, so I aligned the comment and pointed it at runbook step 4.

Doc/comment only. CI green on `459e99f`. This should be the last cycle — re-tagging for a final confirm.

@claude

PR #318, opened 2026-08-01
feat: inference-run observability (variant-selection plot, feature guard, per-band plot) + candidate frequency axis

### LINKED ISSUE #317: [ENHANCEMENT]: Inference-run observability + candidate spectrogram frequency axis
Maintainer-directed (v1.0.0 post-release cleanup session). A batch of inference-run observability + candidate-visualization improvements, none of which change any science numerics:

- [ ] **Candidate spectrogram frequency axis.** Replace the `low → high MHz` text-and-arrow span x-label on candidate/stamp waterfalls with border tick labels (the MHz value at the left and right edges) under a `Frequency (MHz)` axis title, so frequencies are always explicitly listed.
- [ ] **RF variant-selection plot.** The #282 8-variant sweep's winner is surfaced only as a log line + DB scalars — add a plot that makes the winner (and the "tie-break toward the simpler variant" story) explicit: recall@FPR per variant with the winner flagged, the recall traded away shaded, plus AUC/Brier/ECE/n_features companions.
- [ ] **RF feature-count guard.** Same-width latent variants (`z`/`z_mean`/`z_aug`) mean a config↔weights mismatch (mixed `--config-path`/`--rf-path` or a hand-edited config) would silently score the wrong representation — add a load-time assertion that the loaded forest's `n_features_in_` matches the config's declared variant.
- [ ] **Per-band inference-performance plot.** Auto-fire a plot after each inference run (alongside `benchmark_report`) that groups per-cadence preprocessing wall-clock by band (L/S/C/X) and frequency, to reveal whether a band/region is significantly slower.

Docs for the variant-selection plot + the guard will be reconciled in the documentation-sweep PR.

zachtheyek:
Four inference-run observability / candidate-viz improvements. **No science numerics change** — every item is additive or a fail-loud guard.

### 1. Candidate spectrogram frequency axis (`candidate_figures.py`)
The `low → high MHz` text-and-arrow span x-label on the candidate/stamp waterfalls is replaced with border tick labels — the MHz value at the left and right edges — under a `Frequency (MHz)` x-axis title. `render_candidate_figure` now routes its waterfall through the shared `draw_cadence_strip` helper, so all four real-MHz waterfalls share one implementation; bin-order semantics preserved (a negative `foff` prints borders high → low).

### 2. RF latent-variant selection plot (`train.py`)
`plot_rf_latent_variant_selection` makes the #282 winner explicit (previously only a log line + DB scalars): a recall@FPR bar chart per variant ordered simple→complex, the winner flagged, and — when the minimum-margin tie-break passed over a higher-recall variant — a shaded band showing the recall deliberately traded for the simpler representation. Companion panels: AUC / Brier / ECE / feature-count. Reads the in-memory `variant_metrics` (no DB round-trip, no TF); called inline after selection, guarded best-effort.

### 3. RF feature-count guard (`inference.py`)
`init_models` now asserts the loaded forest's `n_features_in_` matches the feature count implied by the config's declared `latent_variant`. Same-width variants (`z`/`z_mean`/`z_aug` all = `num_obs*latent_dim`) mean a config↔weights mismatch from mixed `--config-path`/`--rf-path` pairings or a hand-edited config would otherwise silently score the wrong representation — it now fails loud. (`variant_feature_count()` already existed for this but was unused; the sanctioned HF path always ships matching artifacts so it never fires there.)

### 4. Per-band inference-performance plot (`utils/perband_report.py` + `main.py`)
Auto-fires after each inference run (alongside `benchmark_report`, same `benchmark_report_enabled` gate, fully guarded). Groups per-cadence preprocessing wall-clock by band (L/S/C/X) and frequency from `pipeline_stages`, joined to the run's inference catalog via the planner's own first-appearance cadence grouping. Stdlib-only (`sqlite3`+`csv`+matplotlib), no `aetherscan`/TF imports — same self-contained style as `benchmark_report.py`. Documented in `docs/BENCHMARKING.md`.

### Tests
`TestDrawCadenceStripFreqAxis`, `TestRfFeatureLayoutGuard`, `TestLatentVariantSelectionPlot`, and `tests/unit/test_perband_report.py` (synthetic DB + catalog, umbrella-vs-child isolation, count-guard skip).

Docs for items 2–3 will be folded into the forthcoming documentation-sweep PR.

Closes #317

### DIFF
--- a/docs/BENCHMARKING.md
+++ b/docs/BENCHMARKING.md
@@ -166,6 +166,38 @@ Thresholds are module constants (`DATA_GEN_ROUND_FRACTION`, `GPU_UTIL_INPUT_BOUN
 `PREPROCESS_WALL_FRACTION`, `RAM_PEAK_WARN`) — tune them there if a rule is too eager for your
 hardware.
 
+## The per-band inference plot (`utils/perband_report.py`)
+
+A sibling tool (same stdlib `sqlite3` + `csv` + numpy + matplotlib, no `aetherscan` imports) that
+answers a question the flame timeline can't: **is a whole observing band or frequency region
+systematically slower to preprocess?** It writes
+`{output_path}/plots/perband_inference_perf_{tag}.png` — a two-panel, log-y figure of per-cadence
+energy-detection preprocessing wall-clock: Panel A is a per-band boxplot + jittered strip (bands
+ordered L, S, C, X, each annotated with median/max/n), Panel B scatters the same walls against the
+catalog `Frequency` (MHz), colored by band. The preprocessing wall per cadence is the umbrella
+`inference.preprocess_cadence_<N>` span (its `.read_ed`/`.dedup`/`.extract` children are excluded),
+and the band/frequency come from the run's inference catalog CSV. It fires **automatically at the
+tail of every streaming-CSV `inference` run** right after the benchmark report
+(`_post_perband_report` in [`main.py`](../src/aetherscan/main.py), same
+`monitor.benchmark_report_enabled` gate, same Slack upload, fully guarded — a plot failure never
+fails the run), and is also runnable standalone:
+
+```bash
+python utils/perband_report.py --save-tag test --catalog /path/to/catalog.csv \
+    --db-path /path/to/aetherscan.db
+```
+
+**Join caveat.** The umbrella span carries only the 1-based planner index `N`, and no DB table maps
+`N` to a band, so the cadence → band/frequency map is reconstructed from the catalog by mirroring
+`preprocessing.group_observations_from_csv` (group rows by the default
+`cadence_group_by_cols` = `["Target", "Session", "Band", "Cadence ID", "Frequency"]`, keep the
+6-obs valid groups in first-appearance order; the i-th is cadence `N=i`). This **assumes the run
+used the default group-by columns / expected-obs**. A runtime guard compares the mapped catalog
+cadence count to the umbrella-span count and skips the plot (logged warning, never a crash) when
+they disagree — so a resumed run (fewer fresh preprocess spans than catalog cadences) or a
+non-default grouping degrades to no plot rather than a misleading one. The legacy `--test-files`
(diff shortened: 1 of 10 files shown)


zachtheyek:
Thanks — worked all four findings (`5b7d8ac`):

1. **`_group_catalog_csv` contract** — now requires *every* column the planner grouping needs (`Band`, `Frequency`, the h5-path column, and all group-by columns) and warns + skips (returns `None`) when any is missing, matching `group_observations_from_csv`'s all-columns-required contract, instead of silently regrouping on a subset. Added a missing-column test.
2. **`_is_umbrella_stage`** — replaced the `rsplit`/`isdigit` heuristic with the anchored `^inference\.preprocess_cadence_\d+$`, so a future `_<digits>`-suffixed child can't be misclassified.
3. **Candidate no-MHz fallback** — restored an informative label: `draw_cadence_strip` now sets `Frequency (bin)` when `freq_range_mhz is None` (covers the galleries too), rather than leaving the axis blank; test updated. In production v1 the sidecars always carry `fch1`/`foff` so this is the legacy/degenerate path only, but the label beats a blank axis.

4. **`_check_rf_feature_layout` silent `None`-skip** — keeping as-is by design, but flagging my reasoning in case you disagree: a *fitted* sklearn forest always exposes `n_features_in_`, so the `None` branch is unreachable for a real loaded RF; if the loaded object somehow isn't a fitted estimator, `predict_proba` fails loudly downstream regardless — so silently skipping the guard can't mask a genuine problem, and the defensive skip avoids introducing a new crash path for an exotic estimator type. I can switch it to a `logger.warning` on `None` if you'd rather have the anomaly surfaced.

zachtheyek:
Fixed the optional docstring nit in `5b?` — `map_catalog_cadences`'s "Returns None" clause now names the full required-column contract, not just Band/Frequency. Round 2 was clean/no-blockers and this was the only remaining (optional) note, so I'm closing the loop here — not re-tagging.

PR #320, opened 2026-08-01
misc: default bench data under {data_path}/bench + document loky sem.* leftovers

### LINKED ISSUE #319: [MISC]: Default benchmark data under {data_path}/bench + document loky sem.* leftovers
Maintainer-directed (v1.0.0 post-release cleanup). Two small housekeeping items:

- [ ] **Benchmark data should default under `{data_path}/bench/`.** The suite doesn't currently place benchmark data consistently — `bench_datagen.py` has a required `--data-dir` (no default) and `bench_input_pipeline.py` defaults to `/tmp`. Default both to `{AETHERSCAN_DATA_PATH}/bench/{datagen,input}` (honoring the same env var the config uses) and update the documented example paths, so future benchmark runs land in one place instead of scattering.
- [ ] **Document the benign `/dev/shm/sem.loky-*` leftovers.** `KNOWN_ISSUES.md` documents leaked `psm_*` Python-SharedMemory blocks after a forced kill, but not the `sem.loky-*` POSIX semaphores that joblib's loky resource-tracker can leave on a non-clean teardown (a library artifact, not a pipeline leak — the pipeline's own SharedMemory lifecycle is creator-only-unlink + ResourceManager-registered + verified). Add the cleanup note.

zachtheyek:
Two small housekeeping changes from the v1.0.0 post-release cleanup.

### Benchmark data defaults under `{data_path}/bench/`
The suite didn't place bulk benchmark data consistently: `bench_datagen.py` required an explicit `--data-dir` (no default) and `bench_input_pipeline.py` defaulted to `$TMPDIR/bench_input_pipeline` (`/tmp`). Both now default to `{AETHERSCAN_DATA_PATH}/bench/{datagen,input}` via a small `_default_data_dir()` helper that reads the same env var the config uses (fallback `/datax/scratch/zachy/data/aetherscan`, mirroring `config.py`), so benchmark data lands in one place. `bench_datagen`'s help now notes the round is deleted unless `--keep`. Example paths in the docstrings, `benchmarks/README.md`, and `docs/BENCHMARKING.md` updated to `/datax/scratch/$USER/data/aetherscan/bench/…` (kept `$USER`, not hardcoded). The results-JSON path (`benchmarks/results/`) is unchanged.

### Document the `sem.loky-*` leftovers
`KNOWN_ISSUES.md` covered leaked `psm_*` SharedMemory blocks after a forced kill but not the `/dev/shm/sem.loky-*` POSIX semaphores joblib's loky resource-tracker leaves on a non-clean teardown (SIGKILL/OOM/container exit). These are a library/transitive artifact, **not** a pipeline leak — the pipeline's own SharedMemory lifecycle is creator-only-unlink, ResourceManager-registered, and atexit/signal-wired. Added a sibling cleanup note (`rm -f /dev/shm/sem.loky-*`).

Closes #319


zachtheyek:
Thanks — all three applied (`897b383`):

1. **Dedup** — hoisted the helper to a public `benchmarks/_common.default_bench_data_dir(sub)`; both `bench_datagen.py` and `bench_input_pipeline.py` now import + use it (they already imported `_common`), so there's no copy and one test covers both.
2. **Fallback test** — the test moved to `tests/unit/test_bench_common.py` and now covers **both** the `AETHERSCAN_DATA_PATH`-honored path and the env-unset **fallback** (the load-bearing parity with `config.py:688`). It runs green (2 passed).
3. **KNOWN_ISSUES** — moved the `sem.loky-*` paragraph + its block below the `kill -9 <pid>` line, so the `psm_*` cleanup block and the kill fallback are adjacent again.

PR #322, opened 2026-08-01
feat: machine-scoped display tag for filenames + plot titles

### LINKED ISSUE #321: [ENHANCEMENT]: Machine-scoped display tag for filenames + plot titles (avoid cross-host collisions)
Maintainer-directed (v1.0.0 post-release cleanup). The run tag is `{command}_{datetime}` (e.g. `inf_20260731_182011`). Two runs of the same command that start in the same second on two different machines resolve to the **same** tag, so their on-disk artifacts collide the moment they share a directory — e.g. weights copied bla0 → blpc3.

**Ask:** introduce a machine-scoped **display tag** `{command}_{machine}_{datetime}` used for every artifact FILENAME/path and every plot title + Slack broadcast, so those no longer collide, and so plots consistently show the machine.

**Hard constraint:** do NOT change any DB value. The tag written to / queried from the database (and `config.checkpoint.save_tag`) stays `{command}_{datetime}`. The display tag is a derived, presentation/filename-only string.

zachtheyek:
Introduces a machine-scoped **display tag** `{command}_{machine}_{datetime}` (e.g. `inf_blpc3_20260731_182011`) used for every artifact filename/path and every plot title + Slack message, so artifacts of two same-second runs on different hosts no longer collide when copied together (e.g. weights bla0 → blpc3).

**The DB tag is unchanged.** `config.checkpoint.save_tag`, every DB write/query, and `resolve_save_tag` all still use the plain `{command}_{datetime}`. The display tag is derived, presentation/filename-only.

### Core
- `display_tag(tag, machine_name)` (new `display_tag.py`, stdlib-only, import-safe everywhere): rewrites a genuine run tag to `{cmd}_{machine}_{datetime}`; returns `round_XX`, `final`, empty, and malformed tags **unchanged** (load-bearing — splicing the machine into `round_05` would break the checkpoint-cleanup regex).
- `get_machine_name()` (new, in `db.py`): the single machine accessor, unifying the two former sources (`get_system_metadata()["machine_name"]` vs raw `socket.gethostname()`).

### Threaded through
Model trio / RF / variant / calibrator / eval / shap joblibs, `config_{tag}.json`, `run_state`, log file, `inference_reference_cloud`, round-data dir, every training + inference viz PNG/GIF, benchmark-report + resource-utilization filenames — and every plot title + Slack string (machine moved out of the `({tag}, {machine})` parenthetical into `({display_tag})`, and added to the sites that lacked it).

### Correctness (join-key invariants)
- **Resume** reconstructs `display_tag(save_tag, local_host)` on the same host that wrote the files → matches. `round_XX`/`final` pass through unchanged, so checkpoint load + cleanup are intact.
- **HF upload** runs on the training host, so it globs the display-tagged local files onto the fixed Hub filenames; the HF commit/tag stays the plain DB tag.
- **Inference load** uses explicit `--encoder-path`/`--rf-path` or fixed HF names (display tag irrelevant); the calibrator-by-`replace` operates on the passed basename.

### Deliberate exception
**umap joblibs stay on the plain tag.** `inference_viz.plot_inference_latent_projection` locates `umap_cadence_…_{train_tag}.joblib` by reconstructing it from the *training* run's DB `save_tag` (from the config JSON), potentially on a **different** host whose machine name isn't recorded there — display-tagging umap would break that cross-host reader. umaps aren't among the copied artifacts, so this doesn't reintroduce the collision.

### Tests
`test_display_tag.py` (15 cases incl. passthrough + non-collision), a resume-invariant test (writer name == reader-reconstructed name; a plain-named pair is rejected for a run tag), and updates to the two existing tests that used real `{cmd}_{datetime}` tags.

### Follow-ups before merge
- Rebase over #318 and apply the same treatment to its two new plots (`plot_rf_latent_variant_selection`, `_post_perband_report`/`utils/perband_report.py`) — they don't exist on this branch (off master).
- Cluster smoke: a scaled train → resume → inference run to confirm the write/resume/load paths end-to-end.

Closes #321

### DIFF
--- a/src/aetherscan/db/__init__.py
+++ b/src/aetherscan/db/__init__.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 from .db import (
     get_db,
+    get_machine_name,
     get_system_metadata,
     init_db,
     merge_db,
@@ -14,6 +15,7 @@ from .db import (
 
 __all__ = [
     "get_db",
+    "get_machine_name",
     "get_system_metadata",
     "init_db",
     "merge_db",
--- a/src/aetherscan/db/db.py
+++ b/src/aetherscan/db/db.py
@@ -97,6 +97,11 @@ _SCHEMA_VERSION = 8
 # round (#277). A spawned child process re-imports the module and rebuilds its own cache, so
 # the PID is always correct.
 _SYSTEM_METADATA_CACHE: str | None = None
+# Per-process cache for get_machine_name(): the machine name is a process-constant too, and the
+# display-tag hot sites now call get_machine_name() several times back-to-back (_model_pair_exists
+# twice, save_models/load_models 3x, every `dtag = ...` line). Memoized off the same metadata as
+# _SYSTEM_METADATA_CACHE; neither is reset in-process, so both survive singleton teardown together.
+_MACHINE_NAME_CACHE: str | None = None
 
 
 def get_system_metadata() -> str:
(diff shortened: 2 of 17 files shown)


zachtheyek:
Thanks — both notes applied (`c2b10b2`):

1. **Memoized `get_machine_name()`** via a module-level `_MACHINE_NAME_CACHE`, mirroring its permanent sibling `_SYSTEM_METADATA_CACHE` (which is itself never reset, and no test patches the hostname — so a never-reset per-process constant is the consistent, safe choice; resetting only this cache couldn't change the value anyway since the metadata read is already cached). The hot `display_tag(tag, get_machine_name())` sites are now genuinely free.
2. **Clearer cross-host `--load-tag` error** — `_resolve_load_tag` now explains that on-disk model filenames are machine-scoped (`vae_encoder_{command}_{machine}_{datetime}.keras`), that it looked for the *local-machine* display tag, and that a run's artifacts copied from another host won't be found by `--load-tag` — use `--encoder-path`/`--rf-path`/`--config-path` or the HuggingFace path (fixed names) for cross-host inference. Control flow unchanged; docstring gained a one-line NOTE. This is the deliberate tradeoff of the refactor, now surfaced instead of a bare "not found."

PR #338, opened 2026-08-02
docs: post-v1.0.0 documentation consolidation + fresh-eyes drift sweep

zachtheyek:
## What

The unified post-v1.0.0 documentation pass. It **folds the four auto-generated `claude-update-docs` carrier PRs** into one reviewed PR, fills the remaining documentation gaps from the doc-issue backlog, and runs a fresh-eyes docs-vs-code drift sweep over the fully-merged master.

Closes #311, Closes #295, Closes #304.

## Folded carrier PRs (close after this merges)

The docs bot opened one PR per merge in the 5-PR cleanup batch; their content is verified and rolled in here, so they should be closed in favour of this PR:

- **#326** → `docs/RELEASE.md` packaging extras (`pytest>=9.0.3` security floor, `streamlit>=1.54`)
- **#328** → `benchmarks/README.md` + `docs/BENCHMARKING.md` bench `--data-dir` defaults
- **#331** → the #318 feature docs across 9 files (per-band plot, RF variant-selection plot, feature-count guard, candidate frequency axis)
- **#332** → `KNOWN_ISSUES.md` + `docs/MODELS.md` pre-#293 artifact-compat facts

(Their paired auto-issues #325/#327/#329/#330 close with them.)

## New documentation (the gaps the bot could not cover)

- **#311** — `docs/DATABASE.md`: `PRAGMA temp_store=MEMORY` (and why `mmap_size` is deliberately unset — SIGBUS on shared nodes).
- **#295** — `SKILL.md` producer bullet: the `RoundDataProducer` also pre-generates the RF dataset while the **last** round trains.
- **#304** — `README.md`: a PyPI pip-install path documenting the **v1.0.0 workaround** (`tf_keras~=2.17.0` + `TF_USE_LEGACY_KERAS=1`), the Blackwell-needs-the-container caveat (`CUDA_ERROR_INVALID_PTX`), and the no-CPU-mode caveat. The workaround steps are marked as removable once the v1.0.1 fix (#323) ships.
- **HF "unsafe" flag** — `SECURITY.md`: ProtectAI flags `vae_encoder.keras` because loading it deserializes the registered custom `Sampling` layer; documented as benign (accept + document).
[...]

PR #340, opened 2026-08-02
fix(deps): pull tf_keras for legacy-Keras mode off-cluster (Closes #323)

### LINKED ISSUE #323: [BUG]: pip install aetherscan==1.0.0 cannot load its own v1.0.0 HF weights (Keras 2/3 mismatch; no tf_keras dep)
Found while validating the pip-install path on blpc3 for #304.

## Symptom
A fresh `pip install aetherscan==1.0.0` in a clean venv resolves and downloads the v1.0.0 HF weights, but a bare inference **fails to load its own encoder**:
```
Error loading encoder: Could not locate class 'Functional'. ...
  'module': 'tf_keras.src.engine.functional'
```

## Root cause
The published `.keras` weights are serialized with **tf_keras (Keras 2)** — the NGC container runs in legacy-Keras mode (`TF_USE_LEGACY_KERAS=1`), so `from tensorflow import keras` resolves to real `tf_keras` there (see the compat note in `src/aetherscan/models/vae.py`). A fresh pip install instead gets **Keras 3.15.1** (`keras._tf_keras.keras` shim) and **no `tf_keras` package**, and Keras 3 cannot deserialize a Keras-2 model. No `keras`/`tf_keras` pin exists in `pyproject.toml` / `requirements-container.txt` / `environment.yml` / `aetherscan.def`.

So the documented "happy path" (`pip install aetherscan==1.0.0` → bare inference → HF weights) is broken on any fresh non-container machine. The container path is unaffected (it already has tf_keras + legacy mode). The v1.0.0 release smoke passed because it verified HF *resolution + download*, not full weight *loading*.

## Verified workaround (current)
```bash
pip install "tf_keras~=2.17.0"
export TF_USE_LEGACY_KERAS=1
```
With these, the encoder loads cleanly (`Functional`, 4.65M params).

## Proposed fix (for the next release)
1. Add `tf_keras~=2.17.0` to the package dependencies (`pyproject.toml`; it's already present in the container).
2. `os.environ.setdefault("TF_USE_LEGACY_KERAS", "1")` at the top of `src/aetherscan/__init__.py`, before anything imports TensorFlow, so `from tensorflow import keras` resolves to `tf_keras` for pip users too (no-op in the container, which already sets it).

This is a packaging fix; it reaches PyPI users only via a patch release (v1.0.1). Gating the fix + release timing on the maintainer.

## Related (not this issue)
- Blackwell: pip-TF 2.17's bundled CUDA 12.3 has no sm_120 kernels → GPU compute fails with `CUDA_ERROR_INVALID_PTX` (container required on Blackwell — already documented).
- No CPU mode: both subcommands hard-exit "requires GPU" with no GPU visible.

zachtheyek:
> [!IMPORTANT]
> **Prepared, not for immediate merge.** This is the v1.0.1 fix for #323. Merging it does not cut a release (master stays `1.0.1.dev0`), but per the post-release plan the **maintainer decides v1.0.1 timing** — hold until then. Once v1.0.1 ships, the README pip-install section (added in #338) moves from the documented workaround to the clean `pip install aetherscan` path.

## Problem (#323)

A clean `pip install aetherscan==1.0.0` (and the conda env) installs and resolves the v1.0.0 HF weights but **cannot load the encoder**:

```
Could not locate class 'Functional'. ... tf_keras.src.engine.functional
```

The released `.keras` weights are **Keras-2 (tf_keras)** format, but no manifest pulls `tf_keras` and TF 2.17 defaults to Keras 3. The **NGC container path is unaffected** — its base image ships `tf_keras` and already sets `TF_USE_LEGACY_KERAS=1` (verified in-container: `tf.keras` resolves to `tf_keras.api._v2.keras`), so this PR's `setdefault` is a no-op there. This PR additionally declares the flag in `aetherscan.def`'s `%environment` so a future base-image swap can't silently drop it.

## Fix

- **`tf_keras==2.17.*`** declared in `pyproject.toml` (the "NGC base provides these implicitly" block) and `environment.yml` — the two base-less install paths. The NGC container already ships `tf_keras 2.17` from its base image (verified in-container: `tf 2.17.0` / `tf_keras 2.17.0`), so `requirements-container.txt` keeps **omitting** it, now documented in its header next to `tensorflow` — this keeps the "requirements-container.txt omits base-provided packages" convention intact rather than double-pinning.
- **`os.environ.setdefault("TF_USE_LEGACY_KERAS", "1")`** at the top of `src/aetherscan/__init__.py`, before any submodule TF import, so `from tensorflow import keras` resolves to `tf_keras`. `setdefault` means an explicit env value (or the container's own export) still wins.

## Note on manifest placement

HANDOFF-era guidance said "add `tf_keras` to all three manifests," but doing so would falsify pyproject's own "(so requirements-container.txt omits them)" comment and double-pin a base-provided package (like `tensorflow`, which `requirements-container.txt` also omits). I took the convention-consistent route above; happy to pin it in `requirements-container.txt` too if you'd rather have it explicit there.

## Testing

CI now runs the suite under **legacy Keras** (the pipeline's real runtime), since the `setdefault` fires on `import aetherscan` — a stronger check than the previous implicit-Keras-3 CI. Verified locally (TF-free): `import aetherscan` sets the flag when unset and preserves an explicit `TF_USE_LEGACY_KERAS=0`.

Closes #323.

### DIFF
--- a/.github/workflows/claude-dependency-check.yml
+++ b/.github/workflows/claude-dependency-check.yml
@@ -119,8 +119,8 @@ jobs:
                  `environment.yml` and `requirements-container.txt` (numpy, scipy, scikit-learn,
                  scikit-image, astropy, matplotlib, joblib, setuptools, setigen, imageio,
                  umap-learn, shap, slack-sdk, python-dotenv). Conda-only deps (h5py, hdf5plugin,
-                 psutil, pytest, ipykernel) are intentionally ABSENT from requirements-container.txt
-                 because the NGC base ships them — that is NOT drift.
+                 psutil, pytest, ipykernel, tf_keras) are intentionally ABSENT from
+                 requirements-container.txt because the NGC base ships them — that is NOT drift.
                - The **runtime stack** version must agree across `aetherscan.def` (its `From:` base
                  image tag and `%labels` for Base / Python / TF / CUDA / cuDNN / NCCL),
                  `Dockerfile` (its `FROM ...@sha256:<digest>` — MUST pin the SAME NGC base image,
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,17 +46,21 @@ jobs:
 
       # Mirror the NGC container's dependency surface on a CPU-only runner:
       # tensorflow-cpu stands in for the container's GPU TF 2.17 build, the pinned
-      # extras come from requirements-container.txt, and tf_keras/h5py/hdf5plugin/
-      # pandas/psutil/pytest are installed explicitly because the NGC base image ships
-      # them (so the requirements file intentionally omits them; hdf5plugin is
-      # pinned explicitly in environment.yml too — don't rely on setigen pulling
-      # it transitively). tf_keras backs legacy-Keras mode (TF_USE_LEGACY_KERAS is set
-      # in aetherscan/__init__.py), so the suite runs the way the pipeline does; without
-      # it, importing aetherscan under that flag fails `from tensorflow import keras`.
+      # extras come from requirements-container.txt, and h5py/hdf5plugin/pandas/
+      # psutil/pytest are installed explicitly because the NGC base image ships them
+      # (so the requirements file intentionally omits them; hdf5plugin is pinned
+      # explicitly in environment.yml too — don't rely on setigen pulling it
+      # transitively). tf_keras backs legacy-Keras mode (TF_USE_LEGACY_KERAS) and is
+      # installed separately with --no-deps: pulling it normally would drag in the
+      # 601 MB GPU tensorflow wheel on top of tensorflow-cpu, because its
+      # `tensorflow>=2.17` edge isn't satisfied by the -cpu distribution name. tf_keras
+      # imports the `tensorflow` module tensorflow-cpu already provides — the same
+      # arrangement as the NGC image, where tf_keras sits next to a TF it never installed.
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install "tensorflow-cpu==2.17.*" "tf_keras==2.17.*" -r requirements-container.txt h5py hdf5plugin pandas psutil pytest
+          pip install "tensorflow-cpu==2.17.*" -r requirements-container.txt h5py hdf5plugin pandas psutil pytest
+          pip install --no-deps "tf_keras==2.17.*"
 
       # GPU- and cluster-marked tests need physical GPUs / cluster-resident data;
       # everything else (including slow CPU TF graph tests) runs here
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,15 +46,17 @@ jobs:
 
       # Mirror the NGC container's dependency surface on a CPU-only runner:
       # tensorflow-cpu stands in for the container's GPU TF 2.17 build, the pinned
-      # extras come from requirements-container.txt, and h5py/hdf5plugin/pandas/
-      # psutil/pytest are installed explicitly because the NGC base image ships
+      # extras come from requirements-container.txt, and tf_keras/h5py/hdf5plugin/
+      # pandas/psutil/pytest are installed explicitly because the NGC base image ships
       # them (so the requirements file intentionally omits them; hdf5plugin is
       # pinned explicitly in environment.yml too — don't rely on setigen pulling
-      # it transitively)
+      # it transitively). tf_keras backs legacy-Keras mode (TF_USE_LEGACY_KERAS is set
+      # in aetherscan/__init__.py), so the suite runs the way the pipeline does; without
+      # it, importing aetherscan under that flag fails `from tensorflow import keras`.
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install "tensorflow-cpu==2.17.*" -r requirements-container.txt h5py hdf5plugin pandas psutil pytest
+          pip install "tensorflow-cpu==2.17.*" "tf_keras==2.17.*" -r requirements-container.txt h5py hdf5plugin pandas psutil pytest
 
       # GPU- and cluster-marked tests need physical GPUs / cluster-resident data;
       # everything else (including slow CPU TF graph tests) runs here
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,9 @@ RUN pip install --no-cache-dir -r /opt/aetherscan-requirements.txt \
     && rm -f /opt/aetherscan-requirements.txt
 
 # Mirrors aetherscan.def's %environment: don't pick up host-side Python packages that would
-# shadow the pinned image versions; quieten TF's C++ INFO flood (INFO+ still prints).
+# shadow the pinned image versions; quieten TF's C++ INFO flood (INFO+ still prints); run legacy
+# Keras (#323) — the NGC base already sets TF_USE_LEGACY_KERAS, but declare it explicitly so a
+# future base swap can't silently drop it (aetherscan/__init__.py is the Python-level backstop).
 ENV PYTHONNOUSERSITE=1 \
-    TF_CPP_MIN_LOG_LEVEL=1
+    TF_CPP_MIN_LOG_LEVEL=1 \
+    TF_USE_LEGACY_KERAS=1
--- a/aetherscan.def
+++ b/aetherscan.def
@@ -43,6 +43,11 @@ From: nvcr.io/nvidia/tensorflow:25.02-tf2-py3@sha256:c83b37d26f19ab00d8a13cf974e
     export PYTHONNOUSERSITE=1
     # Quieten TF startup noise; INFO+ still prints, only the C++ INFO floodgate closes.
     export TF_CPP_MIN_LOG_LEVEL=1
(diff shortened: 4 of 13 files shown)


zachtheyek:
Thanks — sharp catches. All three resolved in b78f1a4; the branch is rebased onto the merged docs PR (#338).

**1. CI red — fixed.** Exactly your diagnosis: `requirements-container.txt` correctly omits the base-provided `tf_keras`, so `tests.yml` (which installs from it + explicit extras) had `TF_USE_LEGACY_KERAS=1` set but no `tf_keras`. Added `tf_keras==2.17.*` to that install line — same explicit-extras group as `h5py`/`hdf5plugin`/`pandas`/`psutil` — and updated the `TESTING.md` mirror (command + prose). This is the "omit base-provided ⇒ tests.yml must name each one" invariant biting, as you noted.

**2. Container premise — settled, and it's a no-op there.** Ran your command in-container (bla0):
```
TF_USE_LEGACY_KERAS= 1
tf.keras= tf_keras.api._v2.keras
```
So the container is *already* in legacy mode — the NGC base image sets the flag (not `run_container.sh`/`aetherscan.def`, as you found), and the `setdefault` is a no-op there. That means the `models/vae.py:23-27` comment was stale (it described the flag-*unset* `keras._tf_keras.keras` shim), so I corrected it, and fixed the PR body's "wrapper exports it" wording. Per your suggestion I also added `export TF_USE_LEGACY_KERAS=1` to `aetherscan.def`'s `%environment`, so the container no longer relies on an undeclared base-image ENV a swap could drop (`__init__.py`'s setdefault remains the Python-level backstop).

**3. Unit test — added.** `tests/unit/test_legacy_keras_env.py`: two subprocess assertions (unset → `import aetherscan` yields `"1"`; `TF_USE_LEGACY_KERAS=0` preserved), following the `test_train_distribution.py` shell-out precedent. Verified locally.

**Minor.** Added `tf_keras` to `docs/RELEASE.md`'s base-provided enumeration; switched the pin to `tf_keras==2.17.*` to match `tensorflow`'s style; kept the omit-from-`requirements-container.txt` placement (agreed).

This PR stays **prepared, not for immediate merge** — the maintainer decides v1.0.1 timing.

@claude please take another pass.

zachtheyek:
All landed in fb25bdf.

**1. `--no-deps` for tf_keras.** Confirmed the double-install from the job log — `tf-keras`'s `tensorflow>=2.17` edge isn't satisfied by `tensorflow-cpu`'s name, so pip pulled the 601 MB GPU wheel too. Split it onto its own line: `pip install --no-deps "tf_keras==2.17.*"`, keeping the CPU-only wheel as the single TF. Mirrored in `docs/TESTING.md` (command + prose), and the "tensorflow-cpu stands in" claim is accurate again.

**2. Order-proof legacy flag in conftest.** Good catch — the `__init__.py` setdefault only bit by luck of collection order (the TF-first modules were saved by an alphabetically-earlier `aetherscan` import), so a reordering could have gone green with tf_keras missing. Added `os.environ.setdefault("TF_USE_LEGACY_KERAS", "1")` to `tests/conftest.py` next to the `MPLBACKEND`/`TF_CPP_MIN_LOG_LEVEL` ones, so both a single-module run and the full suite run legacy. Listed it in `TESTING.md:203`.

**3. Assert the resolved lineage.** Added `test_tensorflow_resolves_to_tf_keras_in_this_session` — `assert "tf_keras" in tf.keras.__name__` (deferred TF import so the module stays TF-free for the subprocess cases). That's the assertion that would have gone red on master.

**4. Subprocess helper.** Hardened to match the `test_train_distribution.py` precedent: `timeout=120`, no `check=True`, and `assert result.returncode == 0, f"stdout:\n{...}\nstderr:\n{...}"` so a failing child prints why.

**5. Docstring.** Fixed — `src` is on the path via `pyproject.toml`'s `[tool.pytest.ini_options] pythonpath = ["src"]`, not a `PYTHONPATH=src` prefix.

**6. `bench_gpu.py:62`** — left as-is (deliberately). It's a no-op in-container (base image sets the flag) and `bench_gpu.py` is documented container-only; the only exposed case is running it off-cluster, and adding an import-order dance to a benchmark for that niche isn't worth it against `__init__.py` + conftest covering the pipeline and the suite. Noting it here rather than fixing.

Still **prepared, not for immediate merge**.

@claude one more pass, please.

zachtheyek:
Thanks — and thanks for confirming the `--no-deps` fix against the actual install logs. Landed the two cheap doc nits and fixed the release-note-feeding one; leaving the other two as you called them.

- **(a)** `docs/TESTING.md` — added `test_legacy_keras_env.py` to the `tests/unit/` tree (2e092af). Left the pre-existing `test_bench_common.py` / `test_display_tag.py` tree gaps for a separate de-drift pass rather than widen this PR.
- **(b)** Added the hand-rolled-env hint to the conftest-consequences bullet: a `Keras cannot be imported` collection error → `pip install "tf_keras==2.17.*"`. Fitting, since that's #323's own error surfacing at collection now that conftest sets the flag unconditionally.
- **(d)** Fixed the PR body's stale `~=2.17.0` → `==2.17.*` (both manifests landed as `==2.17.*`) — matters because `claude-release-notes.yml` feeds the body into the v1.0.1 note draft. Left `README.md`'s `~=2.17.0` as-is (it's the user-facing v1.0.0 workaround, not a manifest).
- **(c)** Leaving `GITHUB_AUTOMATION.md:56-58` — it's a summary that defers to `TESTING.md`, so it isn't wrong; not worth expanding a workflow-catalog row with the full install command.

Good call on the `test_tensorflow_resolves_*` vs opt-out-preservation tension — that's deliberate: the opt-out test guards the `setdefault` mechanism, and the lineage test guards that the suite refuses to validate on a backend the pipeline doesn't ship. Both are correct stances on the knob.

Not re-tagging — we're at LGTM and a–d were cosmetic. This stays **prepared, not for immediate merge**; it's queued for whenever you cut v1.0.1.

zachtheyek:
Resuming this PR for the release: rebased onto current master (clean, no conflicts — 47 commits, including the #365 GHCR image publish and the #369 docs reframe) and added one scope-completing commit, b448863.

**Why the new commit:** #365 introduced the `Dockerfile` (the OCI twin of `aetherscan.def`) *after* this branch was cut. Its `ENV` block documents itself as mirroring `aetherscan.def`'s `%environment` — but this PR adds `TF_USE_LEGACY_KERAS=1` to the `.def`, which would leave the two runtime recipes divergent on merge. b448863 carries the flag into the Dockerfile's `ENV` with the same rationale as the `.def` change (NGC base already sets it; the explicit declaration survives a future base swap; `aetherscan/__init__.py` remains the Python-level backstop).

Consequence worth stating: this makes the PR a Dockerfile-input change, so the next release's `image` job will **rebuild** rather than retag the existing `fp-<hash>` digest. That's the designed behavior for input changes (docs/RELEASE.md §"Version bump with no weights/image change") — resulting image contents are identical in practice (the base already exports the flag), but the fingerprint honesty is the point.

Everything else is unchanged from the converged third-pass state. Requesting a fresh pass over the rebase + b448863: @claude

zachtheyek:
Landed §2 in fcdbbd0; the rest is triage — one deliberate decline (§4) and three items routed to where they belong.

**§2 — done.** `tf_keras` added to the dependency auditor's base-provided parenthetical. Agreed it's the automation-side half of the convention argument: the audit prompt was the one place the "declared in two manifests, absent from the third, on purpose" story wasn't written down.

**§1 — both points taken, no change.** You're right that 59d6392's `requirements-container.txt` header edit had already moved the fingerprint — so b448863 added a second reason for a rebuild that was already going to happen, and my "identical in practice" clause overclaimed given the ranged pip layer re-resolves at build time. The release-time mitigation (pull the new `:vX.Y.Z` image and diff `pip freeze` against the v1.0.0 digest) is going into the release checklist for this cycle rather than this PR.

**§3 — deferred to the docs-unification PR, deliberately.** There's a single unified docs PR queued for this release cycle that consolidates the open docs PRs (#367/#363/#342 et al.) plus exactly this class of post-#365 drift. The `GPU_RUNTIME_GUIDE.md:74` base-swap how-to (missing the Dockerfile's `FROM`) and the two missing `TESTING.md` tree entries go there, where they can be reviewed alongside the other GHCR-reconciliation edits — keeping this PR's docs footprint to what its own code changes require. (The TESTING.md tree gaps were already explicitly deferred once in this PR's earlier loop, for the same reason.)

**§4 — declining, with reasons.** The hard lineage assert is the one signal in CI that would have gone red on master for #323; converting it to a skip on `TF_USE_LEGACY_KERAS != "1"` means an accidentally-exported opt-out (a leaked env var in a runner, a future workflow edit) silently disables the only backend-lineage check we have, which is the exact failure class this PR exists to close. The first two tests treat the opt-out as legitimate *for the setdefault contract*; the third asserts the *suite's own environment* is the supported one — different claims, and I think the asymmetry is correct. Red-with-a-clear-message beats green-with-a-skip here.

**§5 — agreed, and it's on the release checklist.** The runbook's post-publish smoke (`pip install aetherscan==X.Y.Z` + bare inference in a clean venv on an Ampere box) *is* the #323 repro, and this cycle it gets run with exactly that framing: the acceptance check for this fix, not just a packaging sanity pass.

Requesting a final pass on fcdbbd0: @claude

PR #346, opened 2026-08-02
feat(inference): guard RF latent-variant identity, not just feature count

### LINKED ISSUE #345: Post-v1.0.0 audit: harden RF feature-layout guard against same-width variant confusion (+ code nits)
Surfaced by the post-v1.0.0 retrospective audit.

## Primary: RF feature-layout guard can't detect same-width latent-variant confusion
`InferencePipeline._check_rf_feature_layout` (src/aetherscan/inference.py) is a feature-**count** check only. Its own docstring motivates it with the z/z_mean/z_aug case — but those variants all have identical width (`num_observations*latent_dim`), so a config↔weights mix *among them* passes both the count check and sklearn's predict-time check, and inference would silently score the wrong representation. The guard is fail-safe (never false-trips), so this is a hardening gap, not a live bug — but the docstring overstates what it catches.

**Fix:** stamp the winning variant onto the deployed forest at train time (`aetherscan_latent_variant_`) and add a variant-**identity** check to the guard. Forests trained before this change (incl. the v1.0.0 weights) carry no stamp → the identity check no-ops, exactly like the count check no-ops on a forest without `n_features_in_`. Clarify the docstring to distinguish the two checks.

## Bundled code nits (same audit)
- `utils/benchmark_report.py`: one `str(...).split('.')[0]` left without `maxsplit=1` while the sibling call uses it (cosmetic consistency).
- `requirements-container.txt`: streamlit-rationale comment says `protobuf<6`; streamlit 1.54 actually accepts `<7` (conclusion unaffected — TF 2.17's protobuf 4.x satisfies both).
- `src/aetherscan/display_tag.py`: the machine name is spliced into filenames unsanitized (inert on RFC-1123 hostnames; add defensive sanitization).

Ships with unit tests + a MODELS.md artifact-contract note.

zachtheyek:
Closes #345. Surfaced by the post-v1.0.0 retrospective audit.

## Primary — RF feature-guard variant identity
`InferencePipeline._check_rf_feature_layout` was a feature-**count** check only. Several latent variants share the same width (`z`/`z_mean`/`z_aug` are all `num_observations*latent_dim`), so a config↔weights mix among them passed both the count check and sklearn's own predict-time check — inference would silently score the wrong representation, the exact case the guard's docstring cited as motivation.

- **train:** stamp the winning variant onto the deployed forest (`self.rf_model.model.aetherscan_latent_variant_ = winner`) right after selection, so it rides through every subsequent `joblib.dump`/`load`.
- **inference:** add a variant-**identity** check after the count check. Forests without the stamp (incl. the released v1.0.0 weights) **no-op** it — so the imminent v1.0.0 inference run is unaffected — exactly as the count check no-ops without `n_features_in_`.
- **docstring:** rewritten to distinguish the two complementary checks; MODELS.md artifact row updated.
- **tests:** +3 cases (matching stamp passes; same-width mismatch raises; absent stamp no-ops).

## Bundled audit nits
- `utils/benchmark_report.py`: `split('.')[0]` → `split('.', maxsplit=1)[0]` for consistency with the sibling call (the `len(split(...))` depth calls and the `[-1]` leaf call correctly keep the full split).
- `requirements-container.txt`: streamlit-rationale comment `protobuf<6` → `<7` (streamlit 1.54's actual bound; conclusion unchanged).
- `src/aetherscan/display_tag.py`: sanitize the spliced machine name to filename-safe chars (identity on RFC-1123 hostnames — blpc3/bla0 untouched).

## Verification
ruff 0.15.22 check + format clean; `py_compile` clean; display_tag sanitizer functionally tested (normal hosts untouched, pathological collapsed, non-run-tags unchanged). TF-importing tests deferred to CI (TF-CPU) per the repo's local-dev constraint.


zachtheyek:
Thanks — thorough pass, and the end-to-end trace matches my read. Worked all four findings in `e29ac35`:

**1. Per-variant forests unstamped → fixed at the fit site.** Moved the stamp to right after `clf.fit` (train.py ~3090), so every `random_forest_{tag}_{variant}.joblib` is self-describing — exactly the same-dir/same-tag mispairing source you flagged. **One deliberate divergence from your suggestion:** you said to *also keep* the line-3181 deployment-point stamp; I dropped it (for a comment) instead. Rationale: after fit-site stamping, `self.rf_model.model` **is** `variant_models[winner]` — the same already-stamped object — so line 3181 would be a redundant re-assignment of the identical value. The comment documents the invariant. Happy to add it back as belt-and-suspenders if you'd rather guard against a future fit-loop refactor, but it read as dead-weight today.

**2. `active_dims` hole → closed, not deferred.** Also stamp `aetherscan_active_dims_` at the fit site and added a set-level identity check scoped to `z_mean_logvar_active` (the only variant whose *columns* are active-dim-indexed). Two equal-length sets (`[0,1,2]` vs `[5,6,7]`) now raise. The docstring is rewritten to describe all **three** checks with increasing specificity, so it no longer implies the count+variant pair is jointly exhaustive.

**3. Untested joblib round-trip → pinned.** Added `TestVariantStampPersistence` to the TF-free `test_latent_variants.py`: fit a 2-tree forest, set both stamps, `dump`/`load`, assert survival + `n_features_in_`. Your ~6-line suggestion, in the TF-free home you pointed at.

**4. Nits.** Guard class docstring updated to the three-check story; dropped the redundant `test_absent_variant_stamp_is_a_noop` (it was argument-identical to `test_matching_count_passes`, now commented as the stampless v1.0.0 path); trimmed the `display_tag` comment's `round_XX`-regex claim — you're right that `_`/digits pass through, so it never protected against a `round_9`-shaped hostname; the justification is now path-safety only. Also noted the stamps on the `_{variant}.joblib` MODELS.md row.

**On your testing note:** I couldn't run the pytest suite locally (TF deadlocks on this iCloud mount), but I *did* verify the load-bearing bits standalone — `variant_feature_count('z_mean_logvar_active', 6, 8, [0,1,2]) == 66` (matches the new test) and both stamps surviving a real `joblib.dump`/`load` round-trip. And the `protobuf<7` bound is confirmed from streamlit 1.54.0's PyPI `requires-dist`. CI (TF-CPU) remains the real gate for the +5 guard cases.

@claude

zachtheyek:
Both notes fixed in `699a8d6` (and thanks for confirming the line-3181 drop — object identity holds, so it stays dropped):

**1. Set vs ordered compare.** You're right — `_active_logvar_columns` sorts, so the columns are a function of the active-dim *set*. Changed the check to `sorted(recorded) != sorted(config or [])` so a reordered-but-equivalent config (`[2,0,1]` vs `[0,1,2]`) no longer false-trips. The comment and error message already said 'set', so they're now accurate; added a line noting the sort rationale.

**2. Untested scoping branch.** Added `test_active_dims_drift_ignored_for_nonactive_variant` — a `z_mean` forest with a deliberate active-dims drift must pass, since the check is scoped to `z_mean_logvar_active`. Also added `test_reordered_active_dims_passes` to pin the set-compare from note 1. Guard now has 8 cases.

**3. display_tag comment.** Trimmed it to the accurate story: you're right the prefix rule alone doesn't get you there (a `round_9` host *would* match `pattern.search`, it just never lands in the `checkpoints/` dir the sweep scans). Rather than restate that whole chain in a sanitizer comment, I dropped the tag-disambiguation claim entirely and kept only the path-safety justification, which stands on its own.

ruff 0.15.22 check+format clean, `py_compile` clean, and the set-compare semantics verified standalone. E501 noted as ignore-listed, so the two >100-char guard lines are within repo norms. CI (TF-CPU) is the gate for the now-8 guard cases.

@claude

zachtheyek:
Thanks for the three passes. LGTM confirmed, CI green. Leaving the optional 'sorted multiset' wording tweak per your note (not worth a standalone commit + CI cycle); I'll fold it in if the file's touched again. Merging on the maintainer's behalf.

PR #358, opened 2026-08-03
docs: document rf_shap_clustering_{tag}.joblib in the MODELS.md artifact table

### LINKED ISSUE #357: [DOCUMENTATION]: document rf_shap_clustering_{tag}.joblib in the MODELS.md artifact table
The persisted-artifacts table in docs/MODELS.md documents `rf_eval_artifacts_{tag}.joblib` and `rf_shap_values_{tag}.joblib` but is missing a row for `rf_shap_clustering_{tag}.joblib` (produced by `plot_rf_shap_explanation_clustering` at src/aetherscan/train.py:6780). It caches the SHAP-space UMAP embedding + KMeans labels so the SHAP-clustering plot reproduces without a re-fit. Add the missing row.

zachtheyek:
Closes #357. The MODELS.md persisted-artifacts table documented `rf_eval_artifacts_{tag}.joblib` and `rf_shap_values_{tag}.joblib` but had no row for `rf_shap_clustering_{tag}.joblib`.

Adds a row (placed with the SHAP artifacts, before the `umap_*` row it cross-references) describing it: the cached `{embedding, cluster_labels}` for the SHAP-space clustering figure — a seeded 2D UMAP of the summary SHAP matrix (`n_summary × 48`) + KMeans (`k=4`) labels — kept separate from the raw-latent `umap_*` reducers so re-rendering that plot reproduces the identical layout without a re-fit. Verified against `plot_rf_shap_explanation_clustering` (src/aetherscan/train.py:6780).

Docs-only.


zachtheyek:
Good catches — fixed in `4c2dcde`:

**1 (real).** You're right that `n_summary × 48` contradicts the variant table — the SHAP matrix width is the *winning* variant's `F` (48–96, since `val_features[summary_indices]` is the deployed variant's feature block). Reworded to `n_summary × F` (48 for the default variants, up to 96).

**3a.** Added the invalidation clause the `rf_shap_values`/`umap_*` rows have: 'delete it to force a fresh fit after changing `shap_max_samples_summary`.'

**2.** Folded the sibling gap: `docs/ARCHITECTURE.md`'s `{model_path}/` tree had the exact same omission (listed `rf_eval_artifacts`/`rf_shap_values`/`umap_*` but not `rf_shap_clustering`) — added the line so #357 closes fully rather than leaving a second stale listing.

**3b (code, not docs — flagging, not fixing here).** The cache being keyed on `{tag}` alone with a bare `os.path.exists` hit is a real (if narrow) fragility: a same-tag rerun with a changed `shap_max_samples_summary` would load a cached `embedding` whose row count no longer matches the freshly-recomputed `val_subtype`/`correct` masks. Out of scope for this docs PR — happy to file a separate issue to fingerprint or size-guard that cache if you want it hardened. (The same bare-tag pattern is worth a look on `rf_shap_values` too.)

@claude

zachtheyek:
Fixed in `6c0b3be`:

**3a — corrected.** You're right, my clause was backwards: `rf_shap_values_{tag}.joblib` is keyed on `{tag}` alone too, so deleting only the clustering cache re-fits UMAP/KMeans over the *same* stale SHAP matrix (same seed/rows → same layout) and buys nothing. Reworded to *'delete it **together with `rf_shap_values_{tag}.joblib`** after changing `shap_max_samples_summary` — that SHAP cache is keyed on `{tag}` alone too, so deleting the clustering cache alone would just re-fit over the same stale SHAP matrix.'*

**6776 — fixed.** Corrected the same stale `48` in the code comment the row mirrors (`n_summary × F`, F = the winning variant's feature width). One-line `src/` comment touch, no logic; ruff clean.

**3b — filed as #359.** Took your refined framing: the real hazard is the *split-cache* case (one regenerated, the other not → `IndexError` at ~train.py:6841), not the mutually-stale case. The issue proposes the `len(embedding) == len(shap_values)` size guard / a fingerprint, and notes `rf_shap_values` warrants the same review since nothing in `src/` removes either file.

@claude

zachtheyek:
Thanks for the three passes — each caught a real error (the hardcoded 48, then my backwards invalidation clause). Merging on the maintainer's behalf. Code-hardening tracked in #359.

PR #361, opened 2026-08-03
feat(container): bind+forward $HF_HOME for a scratch HF cache; document APPTAINER build-env

### LINKED ISSUE #360: [FEATURE]: run_container.sh bind+forward $HF_HOME for a scratch HF cache; document APPTAINER build-env in GPU_RUNTIME_GUIDE
Two related container-env improvements.

## 1. run_container.sh: bind + forward `$HF_HOME` when set
Bare inference downloads the released weights from the HF Hub into the HF cache. Today that defaults to `~/.cache/huggingface` in `$HOME`. Pointing `HF_HOME` at scratch (so downloads persist and don't fill home) does **not** work by itself: verified empirically that the container cannot see an unbound `/datax` path (only the explicitly-bound `data/models/outputs` are visible), so `HF_HOME=/datax/scratch/$USER/hf_home` would be set-but-unwritable.

Fix: when `HF_HOME` is set, `run_container.sh` binds it 1:1 and forwards it. Unset (off-cluster) → nothing added, HF falls back to its default. Bound separately from `AETHERSCAN_EXTRA_BINDS` so an inline `EXTRA_BINDS=/datag` can't clobber it. Keeps the host-specific *path* in per-host config (`~/.bashrc`) while the *tooling* stays path-agnostic.

## 2. GPU_RUNTIME_GUIDE.md doc gaps
- The build-env block showed only `SINGULARITY_TMPDIR`/`CACHEDIR`; add the `APPTAINER_*` equivalents (Apptainer/Ampere hosts use those; Apptainer reads them first, falling back to `SINGULARITY_*`).
- Document the new `HF_HOME` bind/forward + the scratch-HF-cache convention in the container-run section.

Docs-only change to GPU_RUNTIME_GUIDE + a small addition to run_container.sh; bash -n clean.

zachtheyek:
Closes #360. (This is the small tooling+docs PR you asked to fold the GPU_RUNTIME_GUIDE omission into — the last one, #358, already merged, so it's a fresh PR.)

## run_container.sh — bind + forward `$HF_HOME`
Bare inference downloads the released weights from the HF Hub. To keep that cache on scratch (persist across runs, don't fill `$HOME`), `HF_HOME` must point somewhere the **container** can write — and I verified the container **cannot** see an unbound `/datax` path (only `data/models/outputs` are bound). So when `HF_HOME` is set, the wrapper now binds it 1:1 and forwards it. Unset (off-cluster) adds nothing → HF falls back to `~/.cache/huggingface`. Bound separately from `AETHERSCAN_EXTRA_BINDS` so an inline `EXTRA_BINDS=/datag` (used for inference) can't clobber it. The empty-array expansion is `set -u`-safe. Keeps the host-specific *path* in `~/.bashrc`; the tooling stays path-agnostic.

## GPU_RUNTIME_GUIDE.md
- Build-env block: added the `APPTAINER_TMPDIR`/`APPTAINER_CACHEDIR` form (only `SINGULARITY_*` was shown; bla0/Ampere runs Apptainer, which reads `APPTAINER_*` first).
- Container-run section: documented the `HF_HOME` bind/forward + the scratch-HF-cache convention.

## Verification
`bash -n` clean; `git diff --check` clean. (shellcheck not available locally; CI covers it if configured.)

**Operational follow-up (not in this PR, done separately):** `/datax/scratch/zachy/hf_home` created on blpc3 + bla0; `export HF_HOME=…` will be added to each `~/.bashrc` **after this merges + the clusters pull** — so `HF_HOME` is never set without its bind (which would break the download).


zachtheyek:
Fixed in `7e7beb5`:

**1 (should-fix) — added the preflight.** You're right this isn't inference-scoped: `HF_HOME` is a global `~/.bashrc` export, so a missing/relative/purged path would abort *every* `run_container.sh` call (train included) with a cryptic mount FATAL. Added a fail-fast guard matching the SIF-not-found style — clear message + `mkdir -p` guidance, and it catches the relative-path case too. Header doc now says `HF_HOME` must be an existing absolute dir. (Bonus: this also hardens the exact rollout hazard my PR body worried about — a clear error instead of cryptic breakage if the dir's ever gone.)

**2 (default design) — keeping explicit opt-in.** Agreed with your read: defaulting `HF_HOME=$MODEL_PATH/hf_home` would make the wrapper opinionated about cache location and silently relocate it for off-cluster default-path runs, which is what I was avoiding. The preflight is the cheaper fix for the same failure mode.

**3a — RELEASE.md pointer added** (points at GPU_RUNTIME_GUIDE + notes the wrapper binds/forwards `HF_HOME`).

**3b — noted, no code change.** Correct that nothing lints shell today (`.pre-commit-config.yaml` has no shell hook, no shellcheck workflow) — I've stopped implying CI covers it; `bash -n`-clean by inspection, and the array idiom is the `set -u`-safe one you confirmed.

@claude

zachtheyek:
Fixed in `1fd7e1c` — all three:

**2 (set-but-empty) — good catch, real one.** `-n ${HF_HOME:-}` treated set-but-empty as unset, but without `--cleanenv` the empty var still reaches the container, and `huggingface_hub` resolves `""` → relative `hub/` → silent download into the bind-mounted worktree. Switched to `-n ${HF_HOME+x}` (test *set*), so a blank `HF_HOME=` (in `.env` or `~/.bashrc`) now hits the actionable preflight error. Verified the branch logic: empty → error path, unset → skipped.

**1 (should-fix) — README un-staled.** The forwarded-vars enumeration (`SLACK_*`, `AETHERSCAN_*`, `HF_TOKEN`) was exhaustive before this PR; added `HF_HOME` + the note that it's also bound 1:1 / must be an existing absolute dir. Also added a commented `HF_HOME` entry to the `.env` example — you're right the loader runs before the guard, so `.env` is a first-class source (nicer than `~/.bashrc` for per-repo setup).

**3 — GPU_RUNTIME_GUIDE now says mkdir first** (`mkdir -p …` before the export; the wrapper fails fast if missing/relative), matching the rollout sequencing.

Left the writability/nested-bind/shell-lint items per your 'considered, not raising'. @claude

zachtheyek:
Landed the should-fix + both nits in `0be1f06`:

**1 (should-fix) — SKILL.md un-staled, mirroring the README edit.** You're right this one matters more: it's what `CLAUDE.md` points agents at, and the failure mode you named (a future agent 'fixing' the omission with an unconditional `--env HF_HOME="${HF_HOME:-}"` — the exact empty-string injection we spent two passes killing) is the real hazard. `SKILL.md:74` now lists `HF_HOME` + the bind/abs-dir note, and its `.env` example has the commented entry.

**2a (error clarity) — quoted the value.** `HF_HOME='$HF_HOME'` → empty prints as `HF_HOME='' must be an existing absolute directory`, so 'it's empty, not mistyped' is self-evident; the `(or unset HF_HOME…)` clause carries the actual fix for that case.

**2b (README .env) — placeholder + spacing.** `/path/to/hf_home` to match the block's other entries (and so an uncommented copy of a real-but-absent path can't hard-fail every wrapper call), plus the missing blank line so it's its own group.

Left the 'considered, not raising' set as-is. Per your 'land #1 before merge' — that's done, and the code was LGTM on `1fd7e1c`, so I'm merging.

PR #365, opened 2026-08-05
feat(release): publish runtime image to GHCR + pull-first run_container.sh

### LINKED ISSUE #364: Publish prebuilt runtime container image to GHCR
### Summary

Publish the Aetherscan runtime container image to GitHub Container Registry (GHCR) as a 4th version-synchronized release object (alongside the git tag/PyPI, HF weights), so clusters pull a prebuilt image instead of each building the `.sif` locally. Maintainer-requested; designed in a working session.

### Motivation

- The container is the **only** runtime that works on Blackwell (sm_120); today every user builds it from `aetherscan.def` (~9 min, per cluster).
- The image is arch-portable and CPU-arch-uniform across the current x86_64 fleet (blpc3 Blackwell + bla0 Ampere), and the Aetherscan code is **bind-mounted, not baked in** — so one published image serves both clusters and rarely needs rebuilding.

### Scope

- **`Dockerfile`** — OCI twin of `aetherscan.def` (same pinned NGC base digest + `requirements-container.txt`); fork-agnostic (clusters `pull docker://…` and self-convert to a native `.sif`); NVIDIA Deep Learning Container License labels.
- **`.github/workflows/publish-image.yml`** — reusable + dispatchable; auth via built-in `GITHUB_TOKEN`; fingerprints image inputs (base digest + requirements) and **retags the existing digest when unchanged**, rebuilding only when they change.
- **`release.yml`** — new `image` job that gates the PyPI publish (validate-only on the TestPyPI dry run).
- **`utils/run_container.sh`** — priority: local `.sif` → pull GHCR (`:v<version>`) → fail + build instructions.
- **Docs** — `docs/RELEASE.md` (4th synchronized object, CD step, "version bump with no weights/image change", "Container image licensing", v1.0.0 backfill), `README.md` (pull-first + caveats), sync-set docs, and the dependency-check now audits the `Dockerfile` base digest.

### Follow-ups after merge

- Backfill the v1.0.0 image: `gh workflow run publish-image.yml -f version=1.0.0 -f ref=v1.0.0`.
- Set the GHCR package public (first push creates it private).

zachtheyek:
Closes #364.

## What

Publishes the Aetherscan runtime container image to GHCR as a **4th version-synchronized release object** (git tag/PyPI, HF weights, and now the GHCR image), and makes `utils/run_container.sh` pull it instead of requiring a local build.

- **`Dockerfile`** (new) — OCI twin of `aetherscan.def`: same NGC base pinned by digest + `requirements-container.txt`. Publishing OCI (not a `.sif`) is deliberate — Apptainer **and** SingularityCE each `pull docker://…` and self-convert to their own native `.sif`, so no fork-specific artifact is shipped. Carries the NVIDIA Deep Learning Container License labels (§2b notice + `org.opencontainers.image.licenses`).
- **`.github/workflows/publish-image.yml`** (new, reusable + dispatchable) — auth via the built-in `GITHUB_TOKEN` (no PAT). Fingerprints the image inputs (base digest + `requirements-container.txt`) and **retags the existing digest when unchanged**, rebuilding only when they change. Actions are SHA-pinned.
- **`release.yml`** — new `image` job that gates the PyPI publish (validate-only on the TestPyPI dry run, so no heavy build there).
- **`utils/run_container.sh`** — acquisition priority: **local `.sif` → pull `ghcr.io/zachtheyek/aetherscan:v<version>` → fail loudly + build**. Tag derived from `pyproject.toml`; failure-point caveats (arch / driver floor / local TF or requirements changes) documented inline.
- **Docs** — `docs/RELEASE.md` (contract now names four objects; CD step; runbook no-retrain re-bless + v1.0.0 backfill; new **Version bump with no weights/image change** and **Container image licensing** sections), `README.md` (pull-first "Get the `.sif`" + caveats), and the base-digest sync set (`SECURITY.md`, `CLAUDE.md`, `pyproject.toml`, `SKILL.md`).
- **`claude-dependency-check.yml`** — now discovers the `Dockerfile` and flags drift if its `FROM` digest diverges from `aetherscan.def`.

## Why

The container is the only runtime that works on Blackwell, and today everyone builds it per-cluster. The image is CPU-arch-uniform across the x86_64 fleet and the Aetherscan code is bind-mounted (not baked in), so one published image serves both clusters and rarely needs rebuilding.

## Validation

`bash -n` on the wrapper; YAML parse on all three workflows; version-extraction tested on dev (`→ latest`) and release (`→ vX.Y.Z`) versions; all pre-commit hooks pass.

## Post-merge follow-ups (tracked in #364)

- Backfill v1.0.0: `gh workflow run publish-image.yml -f version=1.0.0 -f ref=v1.0.0` (needs the workflow on `master` first).
- Set the GHCR package **public** (the first push creates it private).


zachtheyek:
Thanks — all seven notes were on point; addressed in 01bc3d8 (no rebuttals):

1. **`image` not gated on `verify-weights`** — `image` is now `needs: [guard, tests, verify-weights]`, so a skipped bless blocks the image push too (no lone GHCR tag).
2. **`:latest` moved unconditionally → backfill regresses it** — added a `latest` input (default `true` for real releases via `workflow_call`, `false` for a backfill dispatch) and moved `:latest` in a dedicated step; documented the rule in RELEASE.md's backfill section.
3. **Fingerprint only covered `FROM` + requirements** — `FP` now hashes the whole `Dockerfile` + `requirements-container.txt`, so an ENV/LABEL/added-layer edit forces a rebuild. Kept `BASE` for the log + dep-check compare, but switched it to portable `awk` (drops the GNU-only `grep -oP`).
4. **Pull inherits the `TMPDIR`/`CACHEDIR` wall with no guidance** — pull-failure message now points at redirecting `APPTAINER_CACHEDIR`/`TMPDIR` (or `SINGULARITY_*`) to scratch; added a `trap` so an interrupted pull doesn't orphan multi-GB `.pulling.$$` files; `GPU_RUNTIME_GUIDE.md` gains a pull-first + cache-redirect note.
5. **"falls back to a local build" overstates it** — reworded in README and RELEASE.md: it fails with build instructions, it doesn't build.
6. **Version-pinned tag, version-agnostic cache** — a pulled `.sif` now records its tag in `$SIF.pulled-tag`; a checkout wanting a different tag re-pulls (a user-*built* `.sif` has no sidecar and is used as-is). Also `|| true`'d the `awk` so a missing `pyproject.toml` falls through to `:latest` instead of aborting under `set -e`; `.gitignore` covers the sidecar/temp.
7. **Unpinned `# syntax` directive** — dropped it (nothing uses BuildKit-specific syntax, so no functional loss and no rate-limited Docker Hub frontend pull). The `COPY … && rm` I left as-is on purpose — it's byte-parity with `aetherscan.def`'s `%post` and the `rm` is harmless in the image; happy to drop it if you'd prefer the layers differ.

@claude for a second pass.

zachtheyek:
Second-pass follow-ups all fixed in 933fa8d (no rebuttals):

1. **Stale `.pulled-tag` clobbering a local build** — added the mtime guard: the re-pull `elif` now also requires `! "$SIF" -nt "$SIF.pulled-tag"`, so a `.sif` newer than its sidecar (a local build placed over a pulled image) is treated as user-built (priority 1) and never overwritten. Verified both cases (pulled → re-pull eligible; built-over → preserved).
2. **`INT`/`TERM` trap didn't exit** — split into `trap 'rm -f "$tmp"' EXIT` + `trap 'rm -f "$tmp"; exit 130' INT` + `… exit 143 TERM`, so a Ctrl-C'd pull aborts instead of resuming into the failure branch. Also captured the stale-run warning and repeat it right before `exec` — kept warn-and-continue as the default resilience behavior.
3. **Usage header still promised a build fallback** — fixed `run_container.sh:21` to match note 5.
4. **Re-pull is tag-triggered, not content-triggered** — reworded `GPU_RUNTIME_GUIDE.md` to say it re-pulls on every tag change (every version bump, even a digest-identical retag), named the `<sif>.pulled-tag` sidecar, and documented the mtime build-preservation there and in README's "Build locally instead" block.

@claude for a third pass.

zachtheyek:
Third-pass doc fixes in the latest commit (both accepted):

1. **Wrapper header still carried the two corrected claims** — reworded the acquisition-priority comment: re-pull is tag-triggered (every version bump, digest-identical retag included), and a user-built `.sif` is used as-is either because it has no sidecar or is newer than one (mtime). Added the manual-pull caveat here too.
2. **Manual pre-pull produces no sidecar → silently pinned** — README now states plainly that a manual `pull`/`build` writes no `<sif>.pulled-tag`, so the wrapper keeps it across bumps; let `run_container.sh` pull if you want pinned-tag tracking, or `rm` the `.sif` on a bump.

@claude for a fourth pass.

zachtheyek:
Fourth-pass notes fixed in the latest commit — all three were the same `.devN` nuance, doc-only as you called it:

- The re-pull trigger is a change in the **tag string**, so a release-tag checkout re-pulls on every bump, but a `.devN`/master checkout always resolves to the constant `:latest` and keeps whatever it first cached even after `:latest` moves. Reworded the wrapper header and `GPU_RUNTIME_GUIDE.md` to say exactly that (`rm` the `.sif` + sidecar to refresh).
- Extended the wrapper's caveat list to cover a `.devN` checkout whose `requirements-container.txt` changed **upstream** since the last release — `:latest` tracks the newest release, never master HEAD, so a pull hands you the released image, not master's.

Went doc-only rather than adding a force-pull knob, per your steer.

@claude for a fifth pass.

zachtheyek:
Fifth-pass notes all fixed in the latest commit (all four legitimate — two of them are consequences of my earlier fixes, good catch):

1. **No `:latest` until the first post-merge release** — documented in RELEASE.md's backfill section: the v1.0.0 backfill publishes only `:v1.0.0`, so master/`.devN` checkouts pull-fail → build, which is *correct* (master's `requirements-container.txt` has moved past v1.0.0's, incl. the streamlit security bump), and self-heals at v1.0.1. Kept `latest=false` on the backfill rather than serving a stale image to dev checkouts.
2. **SIGKILL-orphaned `.pulling.<pid>`** — added `rm -f "$tmp"` before the pull; the traps can't run on `kill -9`/OOM/power-loss, and both runtimes refuse to overwrite an existing pull target, so PID reuse would otherwise wedge the pull.
3. **Priority-1 summary line** — reworded to cover the no-sidecar / build-over-pull cases it excluded.
4. **Escape hatch** — the build-locally caveat now names `AETHERSCAN_IMAGE` (your own pushed image); `AETHERSCAN_IMAGE_TAG` only swaps one released image for another and can't remedy those cases.

@claude for a sixth pass.

zachtheyek:
Sixth-pass notes all fixed in the latest commit (all three legitimate — the first is a real bug my pass-5 escape-hatch doc exposed):

1. **Sidecar stored the tag, not the full ref** — so `AETHERSCAN_IMAGE=<fork>` would silently no-op over a cached upstream pull (same tag string). Now `.pulled-tag` stores/compares the full `$IMAGE_REF` (repo:tag); a pre-existing tag-only sidecar mismatches exactly once → one benign self-healing re-pull. Filename unchanged, so no `.gitignore`/README churn. Also fixed the `AETHERSCAN_IMAGE` header blurb to cover the re-pull path.
2. **Fingerprint = whole Dockerfile** — corrected the four RELEASE.md spots still describing the narrow `base digest + requirements` pair; a `Dockerfile`-only edit (e.g. the must-preserve NVIDIA labels) does rebuild.
3. **"publishes only `:v1.0.0`"** — reworded to note the backfill also pushes the internal `fp-<hash>` marker tag, and named `fp-<hash>` in the version-bump section so a package browser isn't surprised by it.

@claude for a seventh pass.

zachtheyek:
Seventh-pass notes all fixed in the latest commit (all three legitimate, and consequences of pass-6 fixes not being fully propagated):

1. **Narrow "base digest + requirements" framing in three more files** — propagated the whole-Dockerfile framing to `Dockerfile:13`, `release.yml:224`, and `publish-image.yml:10`; `grep`'d to confirm no other occurrence remains anywhere. `Dockerfile` now states editing a LABEL rebuilds, right above the must-preserve NVIDIA labels.
2. **`AETHERSCAN_IMAGE` ignored when a local `.sif` exists** — code is intentionally right (mtime guard protects local builds), so added the doc clause: to switch registries you must `rm` the local `$SIF` (or point `SIF=` elsewhere), since a user-built `.sif` takes priority.
3. **"Pulled and cached … (latest)"** — now prints the full `$IMAGE_REF`, so it names the registry when `AETHERSCAN_IMAGE` is set (matching the re-pull/failure warnings).

I greped exhaustively this round to stop the propagation ripple. @claude for an eighth pass.

zachtheyek:
Eighth-pass notes all fixed in the latest commit:

1. **Sidecar tag→ref wording** — finished propagating the pass-6 fix to `GPU_RUNTIME_GUIDE.md` and `README.md` (the sidecar stores the full `repo:tag`, so an `AETHERSCAN_IMAGE` change re-pulls too). `grep`'d to confirm no `pinned/pulled tag` wording remains in docs.
2. **Line reflow** — rewrapped the prose lines my propagation edits over-ran (publish-image.yml header + the three RELEASE.md paragraphs) back under the file's wrap.
3. **Maintenance-branch `:latest`** — added the one-sentence note to the fix-forward paragraph: an out-of-order tag push would drag GHCR `:latest` backward, and the fix-forward-only policy prevents it.

The substantive review converged around pass 6; passes 7–8 have been doc-consistency/cosmetic tail, now addressed exhaustively (grep-driven, so the propagation ripple is closed). I'll treat this as converged and merge once tests are green unless a genuinely substantive issue surfaces. @claude for a final pass.

zachtheyek:
Ninth-pass note fixed in the latest commit — and this one was a genuine functional regression I introduced, so glad the loop caught it (not marginal at all):

**`X.Y.Z` guard fired on the validate-only path** — `publish-image.yml`'s `prep` step asserted `X.Y.Z` with no `if:`, so a dispatch TestPyPI dry run (`test_pypi=true` on `master`, version `1.0.1.dev0`) failed the `image` job, and since `publish` needs `image`, the dry-run upload never ran. Pre-PR that dispatch worked (the guard's tag-match is `push`-only). Fixed: the hard failure now only fires on a real publish (`PUSH=true`); a dry run at a `.devN` version emits a `::notice::` and proceeds to compute the fingerprint (nothing downstream uses `$VERSION` on that path). The documented post-merge dry run was already fine (real `X.Y.Z`); this restores the pre-merge `.devN` dry run the `skip-existing` comment contemplates.

@claude for another pass.

zachtheyek:
Thanks — folded in the two cosmetic notes (`publish-image.yml` notice wording; RELEASE.md dry-run chain now lists `image (validate-only)`). Not re-tagging, since you've converged and confirmed these aren't worth a round. Appreciate the SHA-pin verification and the two-workflow trace on the ninth-pass regression. Merging once the suite is green.

PR #376, opened 2026-08-05
ci: skip the contribution check on maintainer-authored PRs

### LINKED ISSUE #373: claude-contribution-check: skip the maintainer's own PRs
The contribution-compliance check exists to enforce workflow rules on incoming contributions. The maintainer's own PRs don't need the automated reminder (and each run costs a runner spin-up + an action invocation), so the job-level \`if:\` guard that already excludes bot actors should additionally skip PRs authored by \`zachtheyek\`. Issue-opened events keep firing regardless of author (the pull_request context is empty there, so the new clause is a no-op for issues).

zachtheyek:
## Summary

Adds one clause to `claude-contribution-check.yml`'s job-level `if:` guard: PRs authored by `zachtheyek` no longer trigger the compliance check. The check exists to enforce workflow rules on incoming contributions — on maintainer PRs it only cost a runner spin-up and an action invocation per PR.

## Semantics

- The workflow fires on `issues: opened` and `pull_request: opened`. On an issues event `github.event.pull_request` is empty, so the new `login != 'zachtheyek'` clause evaluates true there — **issue-opened events keep firing for every human author**, exactly as before.
- The existing bot-actor exclusions are untouched; this composes with them via `&&`.

## Verification

The expression change is minimal and mirrors the established pattern in the same guard. YAML validity is covered by pre-commit's check-yaml; behavior is observable on the very next maintainer PR (this one fires the *old* definition from master, since `pull_request` workflows run from the base branch — the first PR opened *after* this merges is the real test).

Closes #373

### DIFF
--- a/.github/workflows/claude-contribution-check.yml
+++ b/.github/workflows/claude-contribution-check.yml
@@ -21,11 +21,15 @@ jobs:
     # up a runner and claude-code-action would abort with a "non-human actor" error — a red
     # failure on every bot issue. Skip the whole job instead. (allowed_bots is already unset, so
     # no comment is ever posted; this additionally avoids the noisy failed run.) See issue #83.
+    # Also skip PRs authored by the maintainer (#373) — the compliance reminder is for incoming
+    # contributions. Issue-opened events still fire regardless of author: the pull_request
+    # context is empty on an issues event, so the login clause can't match there.
     if: >-
       github.event.issue.user.type != 'Bot' &&
       github.event.pull_request.user.type != 'Bot' &&
       !endsWith(github.actor, '[bot]') &&
-      github.actor != 'claude'
+      github.actor != 'claude' &&
+      github.event.pull_request.user.login != 'zachtheyek'
     runs-on: ubuntu-latest
     # Contribution check shouldn't take more than 5 mins (refine if data shows otherwise)
     timeout-minutes: 5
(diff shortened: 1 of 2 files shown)


zachtheyek:
Doc drift fixed in 4d30f9b; the two policy items are deliberate — confirming both below — and the optional generalization is declined.

**1. GITHUB_AUTOMATION.md — done.** Both spots now name the maintainer-PR skip: the workflow-table row carries the "nor on maintainer-authored PRs (issues still fire)" clause, and the `allowed_bots` gotcha explains the why (#373) plus the mechanism of the issues-side asymmetry. Agreed it was cheaper here than a post-merge issue → assistant-PR round trip.

**2. The removed safety net is a conscious maintainer decision, not a side effect.** This exact change — the one-line author guard, nothing else — is what the maintainer requested in #373's scope. The forget-proofing value you describe is real, but in practice every maintainer PR here is opened through a workflow that creates/links the issue as part of the process (this PR and its siblings included), so the net was catching nothing on that account while costing a runner + action invocation per PR. If a maintainer PR ever does land unlinked, the failure is visible anyway exactly as you note (label sync copies nothing, release-notes context comes up empty), and the cheap deterministic `closingIssuesReferences` step you sketch is a clean follow-up to add at that point — but adding it now would be scope beyond what was asked.

**3. Issues-side asymmetry — confirmed as a choice.** The requested scope was PR-authored-by-maintainer only. Maintainer-opened issues keep the discussion-link nudge on purpose: the discussions convention has no other reminder mechanism, and unlike the PR flow, issue-filing here isn't mediated by a process that enforces it.

**4. `author_association` variant — declining.** The hardcoded login is the requested semantics; an association-based guard (`OWNER`/`MEMBER`) would silently widen the skip to any future collaborator, which is a policy change nobody asked for. At one maintainer, the literal is the more honest expression of intent — your inline caveat says as much.

Requesting a pass over 4d30f9b: @claude

PR #377, opened 2026-08-05
ci: convert claude-update-docs from per-PR-merge to a weekly batch scan

### LINKED ISSUE #374: claude-update-docs: convert from per-PR-merge to a weekly batch scan
Today the docs-reconciliation workflow fires on every PR merge (plus a workflow_dispatch re-scan of a single PR), which produces one docs issue/PR per merged PR — noisy during release prep, and each run only sees one PR's diff.

Convert it to the same shape as the two existing weekly Claude workflows (claude-dependency-check.yml, claude-flaky-test-tracker.yml):

- \`schedule: cron "0 1 * * 1"\` (Mon 01:00 UTC) + \`workflow_dispatch\`, with the same deterministic ISO-week dedup guard so a double-fire never double-files
- The prompt scans **every PR merged since the last run** (one-week window), reconciles the aggregate changes against the current docs, and — only if something actually needs updating — opens **one** issue and tags the assistant handle for a PR
- Keep the existing prompt's structure and all its contracts (assistant-handle rule, PR-links-issue, never the \`claude\` label)
- Raise the timeout — a week of merges is a bigger scan than a single PR

zachtheyek:
## Summary

Converts `claude-update-docs.yml` from firing on every PR merge to the weekly shape shared by `claude-dependency-check.yml` and `claude-flaky-test-tracker.yml`: `cron: "0 1 * * 1"` (Mon 01:00 UTC = 09:00 MYT) + `workflow_dispatch`, a deterministic ISO-week dedup guard, and **at most one issue per run** covering the whole window.

## Design

- **Scan window** — opens at the previous *successful* run of this workflow (any trigger), so failed runs never advance it and a red week gets re-covered by the next green one; falls back to 7 days when no run history exists. Merged PRs (`--base master`) are enumerated deterministically in a shell step; an empty window skips the Claude invocation entirely.
- **Dedup guard** — week-keyed marker `<!-- aetherscan-update-docs week=<ISO-week> -->`, checked across all issue states (a doc issue closed mid-week must still suppress a re-run). The per-merge era's `pr=<N>` markers are shape-disjoint, so history needs no migration.
- **Contract (a)** (cli.py ↔ README CLI Reference) — the regen pre-step now fires when *any* PR in the window touched `cli.py`, and regenerates off current master, which is canonical regardless of which PR drifted it. The deterministic verbatim-embed step is unchanged except for keying on the week marker.
- **Contract (b)** (canonical docs ↔ CLAUDE.md/SKILL.md), the write-access split for `.claude/` (read-only in CI), the docs-only scope boundary, and the assistant-handle follow-up flow are carried over intact. The prompt's one substantive addition: reconcile against the **aggregate end state** of the window — a gap one PR opened and a later PR closed needs nothing.
- **Timeout** 30 → 60 minutes; `permissions` gains `actions: read` for the `gh run list` window computation.

## Verification

- YAML validated; the guard/window/regen steps are plain shell mirroring the two existing weekly workflows' idioms (including the SIGPIPE-safe here-string from the original).
- Behavior check happens on the first scheduled Monday run (or a manual `workflow_dispatch` once merged — scheduled/dispatched workflows run the default-branch file).

Note for reviewers: until this merges, the per-merge behavior stays live — merges during this PR's review may still file per-PR doc issues; that's expected.

Closes #374

### DIFF
--- a/.github/workflows/claude-update-docs.yml
+++ b/.github/workflows/claude-update-docs.yml
@@ -23,7 +23,9 @@ jobs:
       contents: read
       # actions:read backs the window computation (`gh run list` on this workflow's own history)
       actions: read
-      pull-requests: write
+      # read (not write): the pull_request trigger is gone and every PR touch is a read
+      # (`gh pr diff/view/list`); issue filing rides issues:write
+      pull-requests: read
       issues: write
       id-token: write
 
@@ -55,10 +57,16 @@ jobs:
           fi
 
       # Deterministic scan window + merged-PR enumeration. The window opens at the previous
-      # successful run of THIS workflow (whatever its trigger — a scanned window is a scanned
-      # window) and falls back to 7 days when there is none (first weekly run, or a purged run
-      # history). Failed runs never advance the window, so a red week is re-covered by the
-      # next green one. If nothing merged in the window, the Claude step is skipped outright.
+      # successful SCHEDULED run of this workflow and falls back to 7 days when there is none
+      # (first weekly run, or a purged run history). Two invariants keep coverage total:
+      # failed runs never advance the window (a red week is re-covered by the next green one),
+      # and non-scheduled runs never advance it either — a workflow_dispatch, and any run the
+      # week guard above short-circuited, still concludes `success`, so anchoring on it would
+      # silently drop everything merged between the last real scan and that no-op run. A
+      # dispatch may therefore re-scan a stretch a later scheduled run also covers; that's
+      # harmless — the scan reconciles against CURRENT docs (already-fixed gaps produce
+      # nothing) and the week marker prevents double-filing. If nothing merged in the window,
+      # the Claude step is skipped outright.
       - name: Compute scan window and enumerate merged PRs
         id: window
         if: steps.guard.outputs.already_filed != 'true'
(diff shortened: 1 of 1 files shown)


zachtheyek:
Landed 1, 2, 3, and all three actionable bits of 5 in 9ee753f; declining 4 on token mechanics (details below); 5's `concurrency`/stagger notes routed to a follow-up decision.

**1 — fixed with `--event schedule`.** Your Mon→Wed hole was real, and the schedule-only anchor is the right half of the fix precisely because the *other* direction is safe: a dispatch that re-scans a stretch a later scheduled run also covers is harmless, since the scan reconciles against *current* docs (already-fixed gaps produce nothing) and the week marker blocks double-filing. The step comment now records the full invariant — failed runs never advance the window, and non-scheduled/no-op runs never do either — rather than reading as though "successful" was sufficient.

**2 — limit 200 + `::warning` on cap.** Agreed the un-ordered search API makes silent truncation unrecoverable; the warning names exactly that.

**3 — randomized delimiters.** `${RANDOM}${RANDOM}` suffixes on both heredocs.

**4 — declining, because the deterministic version can't actually be built with the tokens this workflow has.** The handle-last design needs the final tagging write to *trigger* `claude.yml` — but every shell step here authenticates as `github.token`, and GITHUB_TOKEN-created events never trigger workflows. That's not incidental: it's the same mechanics the embed step already depends on ("edits via github.token, so it doesn't itself re-trigger any workflow"). A `gh issue comment` or `gh issue edit` from a shell step therefore summons nobody; only the Claude step's App-token issue creation fires the follow-up. So the current ordering — auditor files at the end of its turn, embed appends seconds later, follow-up boots+checkouts for ~a minute before reading the issue — is the only workable arrangement short of adding a PAT secret, which is a bigger decision than this PR should smuggle in. If the race ever bites in practice, that PAT (or having the follow-up re-read the issue after a delay) is the escalation path.

**5 — the three cheap ones landed** (`pull-requests: read`; `+00:00` timestamps — agreed on not finding out empirically whether the search parser honors `Z`; the SIGPIPE comment rewording — you're right about default-shell `bash -e {0}` vs explicit `shell: bash`). The **`concurrency` group and cron staggering** are real but shared with both sibling weeklies — changing one file to a different shape than its siblings is exactly what this PR is trying to end, so those belong in a small follow-up touching all three uniformly if wanted.

Requesting a pass over 9ee753f: @claude

zachtheyek:
All three landed in a17ef0f; one concession on the point-4 mechanics.

**A — comment now tells the truth.** Your Mon 00:00–01:00 counterexample is correct and I'm not engineering it away either — `gh run list` can't see step outcomes, and every unanchorable-guard-skip variant I can construct is worse than a one-hour-slot residual that additionally requires an in-week re-file. The comment now states the three properties precisely (failed: never; dispatch: never, with the dispatch-only/60-day-disable consequence and the cap-warning tripwire named; guard-skipped scheduled: DOES anchor, residual described) — plus the `createdAt`-not-"now" rationale you pointed out was load-bearing and unwritten.

**B — normalized unconditionally.** One `date -u -d "$SINCE"` after both branches; both paths now emit `+00:00`. Thanks for settling the `Z` question empirically rather than leaving it as my hedge.

**On 4 — conceded on mechanics, unchanged in conclusion.** You're right that a second `claude-code-action` step posting the tagging comment fires `issue_comment: created` on the App token — my "impossible without a PAT" was one notch too strong. So the escalation ladder is: live with the ~1-minute-boot ordering (now) → second action step posting the handle comment (one extra invocation per filing week) → PAT. Recorded here for whoever next touches this; still out of scope for this PR.

Requesting what I hope is the final pass on a17ef0f: @claude

zachtheyek:
All three landed in 6140324, closing the loop on a clean review: the residual comment now names the observed ~04:30 cron deferral instead of claiming a one-hour slot (measuring the siblings was the right instinct — thank you), the fallback is 8 days with the deferral-band rationale in a comment, and the header + prompt both say "successful scheduled run". The nominal-vs-observed MYT header line and the `concurrency`/stagger/`pipefail`-idiom items stay with the uniform three-weekly follow-up as agreed. Merging once checks are green — not re-tagging since the review is clean and these are your own suggestions applied verbatim.

PR #378, opened 2026-08-05
feat(config): raise default training scale — 50 rounds, injection SNR window 1–100

### LINKED ISSUE #372: Raise default training scale: num_training_rounds 20→50, injection SNR window → 1–100
Maintainer-requested default changes for the next release, applied coherently across the curriculum-schedule knobs (they interact — the step schedule validates step_easy_rounds + step_hard_rounds == num_training_rounds at runtime):

- \`num_training_rounds\`: 20 → 50
- \`snr_base\`: 10 → 1, \`initial_snr_range\`: 40 → 99 — the easiest round's injection window becomes exactly SNR 1–100 (signals draw uniform in [snr_base, snr_base + range])
- \`final_snr_range\`: 10 → 9 — the hardest window becomes the quietest decade, SNR 1–10
- \`step_easy_rounds\`/\`step_hard_rounds\`: 5/15 → 10/40 (sums to the new round count; step stays opt-in, exponential remains the default schedule)
- \`exponential_decay_rate\` unchanged: the exponential shape is normalized over progress ∈ [0,1], so it's invariant to the round count

Docs that state the old defaults (docs/TRAINING_PIPELINE.md) must move in the same PR. The released weights keep their own config snapshot — these defaults govern *future* runs only.

zachtheyek:
## Summary

Maintainer-requested default changes for the next release, applied **coherently** across the interacting curriculum knobs rather than as two isolated flips:

| Field | Old | New | Why together |
| --- | --- | --- | --- |
| `num_training_rounds` | 20 | 50 | requested |
| `snr_base` | 10 | 1 | signals draw `snr = snr_base + U(0,1)·snr_range`, so the easiest window becomes exactly **SNR 1–100** |
| `initial_snr_range` | 40 | 99 | ditto (10–50 → 1–100) |
| `final_snr_range` | 10 | 9 | the hardest window becomes the quietest decade, **1–10** (maintainer's pick over preserving the old 10-wide window) |
| `step_easy_rounds` / `step_hard_rounds` | 5 / 15 | 10 / 40 | `_calculate_curriculum_snr` **validates easy+hard == num_training_rounds at runtime** — the old split would crash a 50-round `--curriculum-schedule step` run out of the box; 10/40 is the maintainer's pick |
| `exponential_decay_rate` | −3.0 | −3.0 (unchanged) | the exponential schedule is normalized over `progress ∈ [0,1]`, so its shape is invariant to the round count |

## Scope notes

- CLI: no `cli.py` change — validation is generic (positivity + `initial ≥ final` ordering, all satisfied) and argparse help strings don't embed these numeric defaults, so the README CLI Reference is untouched.
- Docs: `docs/TRAINING_PIPELINE.md` updated where it states the old numbers (round-count default, the SNR-window prose, and the `--retain-round-data` disk-size example, ~147 GB × rounds → ~7.35 TB at 50).
- No unit tests assert these defaults (they all set values explicitly); no test change needed.
- The released weights (`train_20260729_152426`) keep their own config snapshot — these defaults govern *future* runs only, and the code-default-vs-released-config divergence is expected.
- Contract note for the release: changing behavior-when-a-flag-is-omitted is part of the CLI compatibility surface — the release PR will weigh this in the SemVer decision.

Closes #372

### DIFF
--- a/docs/DATABASE.md
+++ b/docs/DATABASE.md
@@ -239,7 +239,8 @@ Index: `(tag, round_number, epoch_number, step_number, model_name, timestamp)`
 loads one capture per frame — up to `latent_viz_gif_max_frames` (500) queries with equality
 on the capture key and the run window trailing — and the old timestamp-second shape
 answered each by scanning the tag's whole window (measured 67× slower per frame at 2 M
-rows; ~1.5 h → ~2 s per GIF pass at the ~100 M-row release scale). Key columns lead so the
+rows; ~1.5 h → ~2 s per GIF pass at the ~100 M-row release scale — measured at the
+pre-#372 20-round default; a 50-round run carries ~2.5× the rows). Key columns lead so the
 dashboard's model-less latest-capture lookup walks the index backward instead of sorting
 the partition; write cost is unchanged because `round/epoch/step` grow monotonically like
 `timestamp`, so both shapes append at the B-tree's right edge
@@ -432,8 +433,9 @@ cost a multi-GB teardown RAM spike for invisible detail.
 `get_db_stats()` returns row counts per table, the covered time range, and the database file
 size — **on-demand diagnostics only** since #301: its per-table `COUNT(*)` full scans
 measured ~13 min per cold-cache launch on the 80 GB production DB (re-paid on every
-retry-loop relaunch), so `Database.__init__` no longer calls it and startup logs only the
-O(1) `db_size_bytes` pragma.
+retry-loop relaunch; both figures from a 20-round run — the #372 50-round default grows
+the DB, and that scan, ~2.5×), so `Database.__init__` no longer calls it and startup logs
+only the O(1) `db_size_bytes` pragma.
 
 ## Growth expectations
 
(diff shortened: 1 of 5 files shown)


zachtheyek:
Both actionable items landed in f5fae1a; the behavioral notes are all confirmations of deliberate choices — responses inline.

**Gap (DATABASE.md) — fixed.** All three rules-of-thumb now compute at 50 rounds (~1.85 B `injection_stats`, ~145 k `training_stats`, ~250 M `latent_snapshots`), and the two *measured* figures got exactly the parenthetical treatment you suggested rather than a rewrite — the 80 GB / ~13 min cold-cache pair and the ~100 M-row GIF benchmark are both labeled as 20-round measurements that scale ~2.5× at the new default. Agreed the cold-cache growth is operator-visible; it'll be named in the release notes.

**Minor (test_model_behavior.py) — fixed.** The `_SNRS` comment now attributes the 10–50 span to the *persisted pre-#372 smoke model* and carries the retune instruction (span 1–100, with your suggested example) for the next smoke-model retrain.

**Behavioral note 1 (floor moves in every round) — intentional, not a side effect.** The floor landing at SNR 1 across the whole schedule is inherent to "window 1–100" under this parametrization (a single global `snr_base`; the schedule narrows only the ceiling), and the maintainer picked the endpoints knowingly — including choosing `final_snr_range=9` (hard window exactly 1–10) over the width-preserving alternative. The old floor-10 curriculum never showed the model a near-noise signal; making faint signals present from round 1 is the point of the change.

**Notes 2 + 3 (late-round collapse diagnostics; no automatic gate) — agreed, run-time practice.** This release re-blesses the existing weights, so nothing changes now; for the first 50-round training run, the two TRAINING_PIPELINE.md diagnostics you cite are the watch-list, and a non-zero `--min-val-auc` tripwire is a good call for that run's invocation. Both are operational choices for that run rather than code in this PR.

**Note 4 (RF set widens to uniform 1–100) — agreed on the metric-shift caveat, declining the new issue.** The threshold/AUC movement on the next retrain is a distribution shift, not a regression, and the release notes for that retrain should say so. The `initial-vs-final_snr_range` question on train.py:2859 already carries the repo's designated marker for exactly this (`# NOTE:`), and filing an issue with no new information beyond the existing marker would be a drive-by by this repo's own conventions — it becomes an issue when someone picks it up with a proposal.

**Note 5** — thanks for verifying `dynamic_range=1.0`; agreed that matters at base 1.

On the could-not-verify: `ruff check` + `ruff format --check` are green locally on the diff, and the full unit matrix runs on this PR's CI (3.10/3.11/3.12).

Requesting a pass over f5fae1a: @claude

zachtheyek:
Took two of the three carry-forwards in 1696e85 despite the LGTM — both are cheap and strengthen exactly what this PR is about; the third is a deliberate leave.

**3 — invariant test added.** `TestCoupledDefaultsGuard.test_default_step_schedule_sums_to_round_count`, framed exactly as you put it: an invariant on the shipped *combination*, not a default-value pin, so it doesn't conflict with the tests-set-values-explicitly rationale. A future `num_training_rounds` bump now fails a unit test instead of silently re-breaking the default step run.

**2 — retune note completed.** The `_SNRS` comment now names `_TOLERANCE`/`_CADENCES_PER_SNR` alongside the span retune, with the noise-floor reasoning, so the eventual retrain instruction is self-contained.

**1 — DATABASE.md migration-log figures left as-is, deliberately.** Agreed with your own framing: the v6→v7 entry is a historical record whose numbers are the *pre-index* costs that motivated the index — scaling them to a hypothetical 50-round run would falsify the history. The asymmetry with :242 is now a choice on the record, per this comment.

Verification note: local pytest in this environment currently hangs at startup (a pre-existing local-env issue — found a pile of stuck pytest processes from earlier sessions too), so the invariant test's execution proof is this PR's CI matrix rather than a local run; `ruff check` is green locally on both files. The test is three lines against the config singleton the rest of the file already exercises.

Requesting a final pass over 1696e85: @claude

PR #381, opened 2026-08-05
fix(train): guard the {tag}-keyed SHAP disk caches against split-cache state

### LINKED ISSUE #359: [BUG]: SHAP-clustering / rf_shap_values caches keyed on {tag} alone — split-cache state IndexErrors
Surfaced by the #358 review.

`rf_shap_clustering_{tag}.joblib` (`plot_rf_shap_explanation_clustering`, train.py:6778-6781) and `rf_shap_values_{tag}.joblib` (`_compute_or_load_shap_values`, train.py:3570-3578) are both cached with a bare `os.path.exists` hit keyed on `{tag}` alone — no fingerprint, no size guard. Nothing in `src/` ever removes either file, so a reused tag keeps whatever artifacts happen to be on disk.

**Failure mode (narrow but real):** if the two caches go out of sync — one regenerated, the other not (deleted by hand, or copied between hosts/runs with a different `shap_max_samples_summary`) — `plot_rf_shap_explanation_clustering` builds `correct_type_mask` from the *fresh* `summary_indices` and indexes the *stale* `embedding` (`embedding[correct_type_mask, 0]`, ~train.py:6841), raising `IndexError: boolean index did not match indexed array`. It crashes loudly rather than mis-plotting.

**Suggested guard:** before accepting the clustering cache, assert `len(embedding) == len(shap_values)` (row-count match) and/or fingerprint the cache on `shap_max_samples_summary` / `n_summary`; re-fit on mismatch. The `rf_shap_values` cache warrants the same review.

Docs for both artifacts were corrected in #358; this tracks the underlying code hardening.

zachtheyek:
## Summary

`rf_shap_values_{tag}.joblib` and `rf_shap_clustering_{tag}.joblib` are two disk caches keyed on the tag alone, written at different times. A reused tag, a changed `shap_max_samples_summary`/`shap_max_samples_interaction`, or one file regenerated without the other leaves stale payloads that `plot_rf_shap_explanation_clustering` indexes with masks sized to the *current* summary subset → `IndexError` (or a silently wrong plot).

## Changes

- **`_shap_cache_consistent()`** (new module-level pure helper): shape-only validation of the SHAP-values payload against the artifacts + config in hand — sample counts (`n_summary`/`n_interact` as `min(config, n_val)`), index ranges (`max < n_val`), and the winning variant's feature width.
- **`_compute_or_load_shap_values`**: the on-disk cache is accepted only when it passes the guard; otherwise a WARNING is logged and the payload is recomputed, overwriting the stale file. (The `n_summary`/`n_interact` computation moved above the cache load so the guard can use it — no behavioral change on the compute path.)
- **`plot_rf_shap_explanation_clustering`**: the clustering cache is accepted only when its row count matches the current SHAP values; refit + overwrite otherwise. Combined with the first guard, both split directions are covered (stale SHAP + fresh clustering recomputes SHAP, then the length check catches the now-stale clustering).

## Tests (`tests/unit/test_train_utils.py`)

- `TestShapCacheConsistent` — the validator: matching payload accepted; summary-count, interaction-count, out-of-range-index, and feature-width mismatches rejected.
- `TestComputeOrLoadShapValuesCacheGuard` — wiring: a consistent disk cache is returned without entering the compute path; a stale one triggers recompute-and-overwrite (SHAP pool + explainer stubbed; the derive_rng/dump plumbing runs for real).
- `TestShapClusteringCacheGuard` — a stale clustering joblib is refit and overwritten instead of crashing; a matching one is used without a refit (UMAP/KMeans stubbed — the guard under test decides whether they run at all, and real fits would dominate wall time without testing our logic).

## Verification

`ruff check`/`format` green. Local pytest currently hangs at startup in this dev environment (pre-existing local issue — stuck pytest processes from prior sessions found alongside), so the executable proof for the new tests is this PR's CI matrix; the tests follow the file's established stub-pipeline idioms (`TrainingPipeline.__new__` + attribute stubs).

Closes #359


zachtheyek:
All five landed in 3c40750 — including both "worth addressing" items in full.

**1 — guard is now total.** `_shap_cache_consistent` takes `object`, prechecks `isinstance(dict)` + the full key set + `isinstance(np.ndarray)` on all five array payloads before any shape/index math — junk, older schemas, and wrong-typed values return `False`, never raise. The clustering guard got the same treatment (non-dict / missing-key / non-ndarray → refit).

**2 — the clustering direction is closed exactly, and the SHAP-side gap is now on the record.** The clustering joblib persists the `summary_indices` it was fit against, and acceptance requires `np.array_equal` with the current indices — so a same-`n_val` regeneration under a reused tag refits instead of serving a wrong-but-plausible embedding. Pre-#359 caches (no indices key) self-heal via one refit through the totality precheck. The SHAP-values-side content fingerprint stays a known gap, and the helper's docstring now says so explicitly rather than implying full coverage — agreed that's follow-up territory, not this PR.

**3 — both skipped payloads covered**: `shap_values_logloss` must be `(n_summary, n_features)` and interaction `(n_interact, F, F)`; the summary check tightened to a full shape-tuple comparison while there.

**4 — done**: `n_rows` hoisted; the warning reports the cached-vs-current row counts and names the indices alternative, so a labels-only or indices-only mismatch no longer logs a same-vs-same puzzle.

**5 — pointed error added**: a stale cache with no trained RF loaded raises `RuntimeError` naming the stale path and the two ways out (retrain or delete), instead of an `AttributeError` inside the dump.

**Tests** — added the malformed-payload validator case, a logloss-shape case, the pre-#359-schema refit case, the same-shape-different-indices refit case (the exact scenario from your point 2), a malformed clustering-cache case, and the matching-cache test now also asserts the accepted joblib is byte-stable on disk. On the 300-dpi `savefig` cost: acknowledged; it's the price of exercising the real plot path end-to-end, and the suite's `slow` marker remains available if it ever shows up in CI timings.

Requesting a second pass over 3c40750: @claude

zachtheyek:
All of it landed in 6d0fcbd — including the point-2 correction, taken in full rather than the narrowed-comment fallback.

**Point 2, fixed properly this time.** You're right that the index comparison was self-refuting (and thanks for owning the pass-1 origin — the contradiction survived into my comment verbatim, which is on me for not re-deriving it). The clustering cache now persists a sha256 **content fingerprint** of the exact SHAP matrix the fit consumed (dtype + shape + bytes), and acceptance requires fingerprint equality — which subsumes both the row count and the indices, and closes your latent-variant-flip scenario: same seed, same `n_val`, byte-identical indices, different matrix contents → different fingerprint → refit. The `_shap_cache_consistent` docstring now states the residual precisely: the clustering side is closed *exactly*, but both caches together can still be coherently stale against new same-shape artifacts through the SHAP-side content gap — that's the follow-up-fingerprint territory we already agreed on. Pre-fingerprint caches refit once via the totality precheck.

**RuntimeError — hoisted and message fixed.** Your inline was right that "delete the stale cache" walked straight into the AttributeError the guard replaces. The check now sits above the whole compute path (covering the no-cache entry too, which previously AttributeError'd the same way), and the message says what actually helps: train or load the RF for this tag.

**Totality nit — closed**, since the docstring does promise it outright: the two index arrays must be 1-D numeric ndarrays before `len()`/`np.max()` see them.

**Warning message** — reports embedding and label counts separately.

**RuntimeError test** — added (`match="no trained RF is loaded"`), stale-cache variant; the no-cache path hits the identical guard.

**Test hygiene** — `_pipeline` now returns `(pipeline, shap_data)` instead of grafting `test_summary_indices` onto the instance; the matching-cache test uses sentinel values (`np.full(…, 7.0)`) and asserts the on-disk payload still holds them; `test_same_shape_different_indices_refit` became `test_same_shape_different_content_refit` — your variant-flip case, pinned directly.

**docs/MODELS.md** — the clustering-joblib row documents the new `{embedding, cluster_labels, shap_fingerprint}` schema, and the delete-both-caches instruction is replaced with the self-validation story (that manual step is exactly what this PR retires).

Requesting a pass over 6d0fcbd: @claude

zachtheyek:
Fixed in 2af189d — and the process failure you flagged is mine to own squarely: I offered the CI matrix as executable proof twice without reading a single run of it. The four clustering tests had never executed on any commit of this PR. That's exactly the failure mode a review loop exists to catch, and you caught it while also being straight about your own two passes; appreciated.

**The blocking fix.** Your inline diagnosis was exact: the stub never set `config.checkpoint.save_tag`, and `_training_plots_dir` builds the savefig base from the config (not the `tag` argument), so `os.path.join(..., None)` raised after the guard had already run and logged. One line in the stub fixes all five.

**Executed proof this time, not asserted.** The local pytest hang turned out to be this dev box's AVX-incompatible TF wheel wedging collection — with a minimal `tensorflow` stub preloaded as a pytest plugin, the SHAP surface now runs locally: **15/15 pass in ~4 s**, including all five clustering-guard tests and the new pre-fingerprint-schema case. The only failures in the full file are `TestEncoderTrainedHeuristics` (builds real encoders — genuinely needs real TF; green in CI). CI on 2af189d is the remaining cross-version check, and this time I'll read it before calling anything done.

**Nits, all landed:** `np.integer` (agreed — `np.number` admitting float indices would make `int(np.max(...))` lie rather than raise); `embedding.shape == (n_rows, 2)` for symmetry with the SHAP-side `ndim` gate; `shap_fingerprint` computed once above the block and reused by the dump; and the new `test_pre_fingerprint_schema_with_matching_rows_refit` pins the key-set precheck in isolation — the actual migration path for every clustering joblib on disk today, previously only covered bundled with a row-count mismatch.

Thanks also for the fingerprint verification in pass 3 — the clean-rerun cache-hit property (astype copy → joblib round-trip → ascontiguousarray normalization) was the part I'd argued but not proven, and your derivation closes it.

Requesting a pass over 2af189d: @claude

zachtheyek:
Closing the loop on a clean review: both cosmetics landed in 87da462 (the lower index bound — agreed that silent from-the-end indexing is worse than a raise; and the class docstring now describes the fingerprint-era guard), and the branch is rebased onto current master per the housekeeping note. 15/15 SHAP tests green locally post-change. Merging once CI is green — not re-tagging since the review is clean and these are your own suggestions applied verbatim.

PR #382, opened 2026-08-05
feat(db): schema v9 — round-scoped injection index + covering latent-keys partial index

### LINKED ISSUE #375: Audit db.py's indexes against the real query patterns (db.py + dashboard.py)
All SQL in the codebase lives in exactly two files — src/aetherscan/db/db.py (writes + queries, all CREATE INDEX statements ~lines 305–634) and src/aetherscan/dashboard.py (read-only dashboard queries). The index set has grown organically; audit it deterministically:

- Enumerate every query in both files; extract each one's filter, join, and sort columns
- Check each access pattern against the existing composite indexes (idx_*_filter, idx_injection_stats_by_stat, idx_latent_snapshots_by_key, the idx_inference_results_supersede partial)
- Add missing indexes for real access patterns and drop indexes nothing uses — **prioritizing query performance over disk footprint**
- Ship the reasoning per index (which queries it serves / why it's dead) in the PR

zachtheyek:
## Summary

The #375 index audit: every `CREATE INDEX` in `db.py` checked against every query in `db.py` + `dashboard.py`, via two independent full-inventory analyses cross-checked by adversarial `EXPLAIN QUERY PLAN` verification on production-shaped replicas (no `ANALYZE`, matching production). Performance over disk, per the issue.

## Outcome

**Keep all 9 live indexes.** Each traced to concrete consumers (details in the commit message); the v7-retired shape stays dropped; the idempotent migration-block re-creates stay for pre-v7 upgrade correctness.

**Add `idx_injection_stats_by_round (tag, round_number, timestamp)`.** `round_number` appeared in *no* `injection_stats` index, leaving the two worst uncovered shapes in the audit on the largest table (~190 M rows / 80 GB at catalog scale):
- the training-resume supersede `UPDATE` (blocks the writer thread; whole-partition row fetch → bounded seek; replica-measured 8.5 s → 0.46 s), and
- round-bounded `query_injection_stat_time_span` (row fetch per partition entry → covering index-only scan; 8.8 s → 0.2 s hot).

**With planner guards — the load-bearing subtlety.** Verification proved the naive ADD is a regression: with no `sqlite_stat1`, SQLite's cost model prefers `by_round`'s leading range over `by_stat`'s four-equality prefix, flipping the ~165-query end-of-run plot pass onto whole-round-range row fetches (measured 1.4–10.7× per query; the exact pathology schema v7 shipped `by_stat` to fix). So `query_injection_stat` and `query_injection_stat_stability` write their round terms as `+round_number` (SQLite's unary-plus term disqualification — semantics identical), and stability additionally guards `GROUP BY +round_number ORDER BY +round_number`: empirically, a bare `GROUP BY round_number` re-attracts `by_round` on timestamp-less shapes purely for its free grouping order.

**Add `idx_latent_snapshots_keys` — covering partial (`WHERE superseded = 0`).** `query_latent_snapshot_keys`' bare `superseded = 0` proves the partial predicate; after the tag equality the column order matches its `DISTINCT` projection and `ORDER BY` prefix exactly, so the once-per-GIF-pass key enumeration becomes an index-only scan instead of a whole-partition row fetch dragging `latent_vector` JSON payloads (~2 M rows/tag, ~100 M at release scale). `include_superseded=True` audit calls fall back to `by_key` — accepted.

**Rejected: a `system_resources (tag, resource_type, resource_name, timestamp)` candidate** for the teardown decimation pass — verified the planner never chooses it unforced for *any* live query (3/3 refutations), so it would be pure write-cost.

## Migration

- `by_round` follows the v7 convention (CREATE in `_init_database`, re-executed in the v9 block); the latent partial follows the v8 convention (block-only — its predicate needs the v1-ALTER `superseded` column). `_SCHEMA_VERSION` 8 → 9.
- **Operator-visible**: upgrading an existing catalog-scale DB builds `by_round` via a one-time full-table scan + sort — the first launch stalls in `_migrate_schema` for minutes to tens of minutes. Documented in `DATABASE.md`'s migration table + prose.

## Tests & verification

- `_EXPECTED_INDEXES` updated — its equality assertions make the existing fresh-DB and v0→current migration tests cover the new final state automatically.
- New `TestV9IndexPlannerContracts`: `EXPLAIN QUERY PLAN` pins **both directions** — the supersede/time-span shapes *use* `by_round` (covering for the aggregate), the guarded plot/stability shapes *stay off* it, and the latent-keys shape is a covering scan on the partial index.
- Ran the full contract set + builder-level result correctness (guarded SQL returns identical rows/aggregates) against the real `Database` init/migration path locally on sqlite 3.50.4 — all green. (Local *pytest* remains broken in this dev env — TF wheel needs AVX, unavailable under Rosetta — so the pytest execution proof is this PR's CI matrix; the local run exercised the same assertions directly.)
- One merge-order note: this PR and #378 both touch nearby `docs/DATABASE.md` text; whichever merges second may need a trivial rebase.

Closes #375

### DIFF
--- a/docs/DATABASE.md
+++ b/docs/DATABASE.md
@@ -342,7 +342,7 @@ attempt, each with its own span.
 | v6 | `is_finite INTEGER DEFAULT 1` on `training_stats` (#289 NaN-write hardening) | additive `ALTER TABLE ... ADD COLUMN` |
 | v7 | the index sweep: the `idx_injection_stats_by_stat` secondary index on `injection_stats`; `latent_snapshots`' index reshaped to `idx_latent_snapshots_by_key` | `DROP INDEX IF EXISTS idx_latent_snapshots_filter` in the migration block; the CREATEs run in `_init_database()` and are re-executed there |
 | v8 | the `idx_inference_results_supersede` partial index on `inference_results` (`(tag, npy_path) WHERE superseded = 0`, #301) | the `if version < 8` block creates it — deliberately NOT `_init_database()`: the partial predicate needs the `superseded` column the v1 ALTER adds, so an init-time CREATE would fail on a pre-v1 file. Fresh databases reach the block too (version 0 → current) |
-| v9 | the #375 index audit: `idx_injection_stats_by_round` (`(tag, round_number, timestamp)`) + the `idx_latent_snapshots_keys` covering partial index; `+round_number` planner guards in the two injection query builders | `by_round`'s CREATE runs in `_init_database()` and is re-executed in the block (v7 convention); the latent partial lives only in the block (v8 convention — its predicate needs the v1 `superseded` column). **Upgrading an existing catalog-scale DB (~190 M `injection_stats` rows) builds `by_round` with a one-time full-table scan + sort — expect the first launch to stall for minutes to tens of minutes, and budget transient WAL disk headroom well beyond the final index size (the build's pages accumulate in the `-wal` file until checkpoint)** |
+| v9 | the #375 index audit: `idx_injection_stats_by_round` (`(tag, round_number, timestamp)`) + the `idx_latent_snapshots_keys` covering partial index; `+round_number` planner guards in the two injection query builders | `by_round`'s CREATE runs in `_init_database()` and is re-executed in the block (v7 convention); the latent partial lives only in the block (v8 convention — its predicate needs the v1 `superseded` column). **Upgrading an existing catalog-scale DB builds both indexes in one stall — `by_round` over ~190 M `injection_stats` rows and the latent partial over the live subset of ~100 M `latent_snapshots` rows, each a one-time full-table scan + sort — expect the first launch to stall for minutes to tens of minutes, and budget transient WAL disk headroom well beyond the final index sizes (the builds' pages accumulate in the `-wal` file until checkpoint)** |
 
 - **v0 → v1**: `ALTER TABLE ... ADD COLUMN superseded INTEGER DEFAULT 0` on the four tables
   above — the only in-place change SQLite supports is additive `ADD COLUMN`, which is exactly
@@ -402,9 +402,10 @@ attempt, each with its own span.
   `+round_number` planner guards in the two injection builders ship in the same change —
   without them the no-`ANALYZE` cost model moves the ~165-query plot pass onto `by_round`
   (measured 1.4–10.7× per-query regressions on production-shaped replicas), resurrecting
-  the pre-v7 pathology. Upgrade cost: building `by_round` over an existing catalog-scale
-  DB is a one-time full-table scan + sort — the first launch after upgrading stalls in
-  `_migrate_schema` for minutes to tens of minutes; later launches are unaffected.
+  the pre-v7 pathology. Upgrade cost: building `by_round` (and the latent partial) over an
+  existing catalog-scale DB is a one-time full-table scan + sort per index — the first
+  launch after upgrading stalls in `_migrate_schema` for minutes to tens of minutes; later
+  launches are unaffected.
 
 Fresh databases get the full current schema from the CREATE statements and are just stamped
 (the block-only partial indexes — v8's supersede index and v9's latent-keys index — land via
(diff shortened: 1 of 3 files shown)


zachtheyek:
All three main findings and every minor landed in 47ddfe1 — finding 1 in full, and finding 2 settled empirically rather than by assertion alone.

**1 — the tests now pin the builders, not SQLite.** You were exactly right that the literals proved nothing about `db.py`. `TestV9IndexPlannerContracts` now installs a trace on `Database._get_connection`, drives the *real* APIs (`query_injection_stat`, `query_injection_stat_stability`, `query_injection_stat_time_span`, `query_latent_snapshot_keys`, `query_latent_snapshots`, and both `mark_superseded` shapes via the inline no-writer path), captures the emitted SQL, and EXPLAINs that. The guarded tests additionally assert `+round_number` (and `GROUP BY +round_number`) appear in the captured statement — so deleting a guard now fails on the string *and* on the plan flip. Positive complements added throughout: the guarded shapes must *keep* `idx_injection_stats_by_stat`, not merely avoid `by_round`.

**2 — settled with plans, then pinned.** Ran both newly-competing shapes through the real APIs against the real schema locally (sqlite 3.50.4, no ANALYZE): the per-frame `query_latent_snapshots` stays on `by_key`'s six-column seek, and the latent `mark_superseded` UPDATE stays on `by_key (tag=?, round_number>?)` — the partial never gets picked despite predicate eligibility. Both are now standing contract tests (`test_latent_per_frame_query_keeps_by_key`, `test_latent_supersede_update_keeps_by_key`), so your "don't rely on it without ANALYZE" point is enforced rather than argued. Your `_ALIVE`-immunity observation for the dashboard matches the audit's own notes.

**3 — operator UX.** A `logger.info` now fires *before* the v9 `by_round` build ("…can take minutes to tens of minutes; do not interrupt"), and both the migration-block comment and `DATABASE.md`'s v9 rows carry the transient-WAL-headroom warning. The `temp_store=MEMORY` RSS question is a fair "measure on the replica before the first real upgrade" item — noted for the upgrade run rather than answered from here.

**Minors:** the affinity boundary is documented in the guard comment ("semantics unchanged for the int-typed params this API takes", with the string-coercion trap named); `DATABASE.md:409` generalized to the two block-only partials; the `_init_database` latent CREATE now points at the block-only v9 sibling.

**Write-cost symmetry — measured, since you asked.** 1M-row bulk insert at the production chunk size (50k/commit) on a WAL+NORMAL replica: 233k rows/s with the two pre-v9 indexes → 225k rows/s with `by_round`, i.e. **~3.5%** on the hottest write path — the append-ordered maintenance the cost-lens verification predicted (tag constant, round/timestamp monotonic), against 18–43× wins on the two target reads. The rejected `system_resources` candidate failed a different test: the planner never *chose* it unforced for any live query, so its (also small) write cost bought literally nothing.

Requesting a pass over 47ddfe1: @claude

zachtheyek:
All of it landed in 6831cee (rebased onto current master while at it — the #385 gap you noted).

**1 — count-pinned, and the stability test now exercises what it claims.** You were right on both sub-points: `query_injection_stat` asserts exactly **2** guarded terms; the stability call now passes `end_round_number` and asserts exactly **4** (`WHERE` start + end, `GROUP BY`, `ORDER BY` — each individually deletable, each now individually pinned, with the `GROUP BY`/`ORDER BY` forms also string-asserted). Your vacuous-pass analysis of the substring checks was exact — the GROUP BY clause alone satisfied the stability `in`-check with the hardest-case guard deleted.

**3 — EXISTS-gated.** Agreed the true-positive audience was one event ever and the false-positive audience was permanent (and Slack-visible). The `SELECT EXISTS(SELECT 1 FROM injection_stats)` probe gates it exactly as you sketched — O(1) at any size, keeps the warning for the genuinely-old v0 upgrade that `version > 0` would have wrongly skipped.

**Minors, all landed:** plan strings now end with `[sqlite <version>]` so a red pin years from now reads as a toolchain change, not a deleted guard; `_plan_of_last` fails by name ("no captured statement contains …, N captured") instead of a bare `StopIteration`; and the latent-keys contract now also asserts **no TEMP B-TREE** — index-only *and* sort-free, closing the one-assertion gap under the DATABASE.md claim.

**Executed proof, this time locally too:** the dev-box pytest hang turned out to be an AVX-incompatible TF wheel wedging collection; with a minimal TF stub preloaded as a pytest plugin, the full `tests/unit/test_db.py` is green locally — **76 passed** — on sqlite 3.50.4, alongside the CI matrix's incidental multi-SQLite coverage you pointed out.

On your trace-fixture audit: thank you for attacking it for vacuous passes rather than accepting it — the PRAGMA-exclusion, expanded-SQL, needle-uniqueness, and dead-writer-inline-path observations are all correct and now effectively documented by your review.

Requesting a pass over 6831cee: @claude

zachtheyek:
Closing out a clean review: c09a0b8 lands the one change you said you'd actually make (the v9 warning — and DATABASE.md's two stall notes — now name both expensive builds, not just by_round) plus the negative-assertion nit (the direct 'no unguarded round_number term reaches the planner' form, which also catches a moved guard). The version-suffix placement stays as-is per your own 'harmless today' read — noted as a boundary if anyone adds a plan.endswith-style assertion. Branch rebased onto current master for real this time (the earlier rebase ran against a stale origin/master ref — your merge-base check caught it). All seven planner contracts green locally post-change. Merging once CI is green — not re-tagging on a clean review.

PR #404, opened 2026-08-08
Inference triage + throughput: frequency filtering, OOD review ordering, completion-order streaming

### LINKED ISSUE #394: Inference viz: candidates-by-frequency plot
Add a figure to the inference visualization suite showing where candidates land in frequency: x = frequency (MHz), y = target (rows sorted by candidate count), dots colored by band, with report-excluded frequency ranges (see the report-time frequency-exclusion filter issue) shaded when configured.

Motivation: run inf_blpc3_20260807_011509 produced 569 candidates with 49% concentrated in two target/bands, and several top candidates sit in known RFI allocations (GPS L1 ~1575.42 MHz, Iridium 1616-1626.5 MHz). Vertical alignment of dots across many targets is the visual signature of multi-target RFI; nothing in the current suite shows it.

Rendered via the existing exception-guarded suite path (inference_viz.render_inference_visualizations), sourced from inference_results (prediction=1) like the candidate gallery.

### LINKED ISSUE #395: Report-time frequency-range exclusion filter for candidates
Add `inference.report_exclude_frequency_ranges` (config) + a repeatable `--report-exclude-frequency-range START_MHZ END_MHZ` CLI flag (default: off). Semantics are strictly report-time: candidates in an excluded range are still detected, still written to inference_results, and their figures still rendered and saved — but they are dropped from the Slack candidate uploads and reported separately in the tallies:

- final run summary log: original candidate count, excluded count, and reported count after exclusion
- inference_summary card: same three numbers
- candidate gallery: built from non-excluded candidates, title noting how many were excluded
- per-candidate figures: rendered and saved for the top-K as today; Slack upload skipped for excluded ones

Validation: each range must be two finite floats with 0 < start < end (MHz); reject anything else at CLI-validation time.

Motivation: in inf_blpc3_20260807_011509, top candidates include GPS L1 (1576.39 MHz) and Iridium (1626.46 MHz) hits — deterministic allocations that waste human review time. A report-time filter (never a detection-time one) keeps the science record complete while cleaning the review surface.

### LINKED ISSUE #396: Standalone candidate RFI-triage report script (multi-target coincidence + known bands)
New `utils/candidate_rfi_report.py` following the utils/perband_report.py pattern (stdlib sqlite3 + numpy, no aetherscan imports, runnable off-cluster against a fetched DB). For a given tag it reads inference_results and reports:

- multi-target frequency coincidence: candidates whose frequency (within a tolerance window) appears across >= N distinct targets — the classic RFI signature, catching transmitters that are in no static list
- known-RFI allocation flags from a built-in band table (GPS L1/L2/L5, GLONASS, Iridium, ...)
- per-target concentration (e.g. NGC1172 179 + MESSIER87 102 = 49% of all candidates in inf_blpc3_20260807_011509)
- the same report-time frequency-exclusion accounting as the pipeline flag, for parity off-cluster

Output: console summary + optional CSV. Flags candidates, never deletes — the DB is untouched.

### LINKED ISSUE #397: Candidate triage: persist reference-cloud latents, OOD scores, gallery ranking tie-break
Three related additions for triaging saturated P=1.000 candidates (569-way confidence tie in inf_blpc3_20260807_011509):

1. Persist the reference-cloud reservoir's latent z_mean rows in inference_reference_cloud_{tag}.npz (they are already in memory at finalize; ~2 MB at default capacity). This gives every run a survey-representative latent sample.
2. Candidate OOD scores, computed end-of-run (exception-guarded like the viz suite): (a) survey-relative — Mahalanobis distance of each candidate's stored latent_vector against the reference cloud's latent distribution, reported as a percentile of the cloud's own distances; (b) training-relative — the same against the training run's true-class latent features from rf_eval_artifacts_{train_tag}.joblib when that artifact is available (graceful skip otherwise). Scores land in a candidate_triage_{tag}.csv beside the plots plus a compact log summary; both are triage surfaces only — no score changes any science decision.
3. Candidate gallery/figure ordering: primary confidence desc (unchanged), tie-broken by survey-OOD distance desc (most unusual first) when available, then MC std asc. Today's order within the saturated tie is arbitrary.

Rationale for OOD-as-triage: a real technosignature is also OOD with respect to synthetic training data, so distances must rank review order, never gate candidacy.

### LINKED ISSUE #398: Mislabeled 'Memory usage' log line prints array nbytes, floors to 0.00 GB
preprocessing.py:1574 (load_inference_data) logs `Memory usage: {cadence_array.nbytes / 1e9:.2f} GB` — the value is the just-built numpy array's byte size, not process memory, and %.2f GB floors anything under ~5 MB to "0.00 GB". In inf_blpc3_20260807_011509 that produced 131 "Memory usage: 0.00 GB" lines (all cadences with <= 24 stamps), which reads like a broken gauge. The real memory monitor (monitor/monitor.py, psutil) is a separate, working path.

Fix: rename to what it measures with a unit that doesn't floor — `Cadence array size: {nbytes/1e6:.1f} MB` — and the same for the training-side twin at preprocessing.py:1255 (`Background array size`). Nothing in the repo parses these lines.

### LINKED ISSUE #399: Flip stamp-cache pruning default to OFF; explicit --prune-stamps opts in
Today inference.prune_stamps=None means AUTO: pruning ON for the fingerprint-scoped default cache dir, OFF when --preprocess-output-dir is set. Consequence: a default-configured subset run deletes every stamp .npy after scoring, so re-scoring the same data (new RF/encoder with the same ED config, threshold sweeps) re-pays the full preprocessing cost — 25.8k stage-seconds in inf_blpc3_20260807_011509 — even though the cache machinery would have skipped it entirely.

Change: default OFF (keep stamps; the cross-run cache works out of the box), with pruning explicitly enabled via the existing --prune-stamps flag. Document the trade-off in the flag help, the config comment, and docs/INFERENCE_PIPELINE.md: keeping the cache buys restart/re-score performance at ~1.1 GB/cadence average disk (380 GB for the 350-cadence subset); pruning bounds disk for catalog-scale runs (a full catalog writes ~TBs, exceeding /datax free space — catalog runs MUST pass --prune-stamps). Log the projected trade-off at run start so a catalog launch without pruning is loud.

### LINKED ISSUE #400: Remove unreachable full-array sanity passes in load_inference_data
preprocessing.py:1555-1557 runs np.min/np.max/np.mean over the full loaded cadence array to gate normalization-failure raises. The raises are mathematically unreachable: rows are validity-filtered upstream (NaN/Inf/non-positive-max rejected in _downsample_cadence, _lognorm_worker, _log_norm_chunk_vectorized) and normalization produces exact-IEEE [0,1] bounds by construction (x - min(x) has min exactly 0.0; x/max(x) has max exactly 1.0; zero-range rows skip division). np.mean feeds one log line. No test asserts on any of it; the [0,1] contract is separately pinned by tests/unit/test_preprocessing.py oracle tests.

Cost: 3 of 13 full-array passes on the streaming hot path — ~400-550 s per subset run of the 2,609 s load_lognorm total (max single-cadence load was 743 s), paid at peak allocation on up-to-65 GB arrays, and directly inflating the prefetch head-of-line stalls.

Change: delete the three reductions and the raise chain; keep a defense-in-depth spot-check of the same three conditions (same raise message) on cadence_array[0] only (~200 KB, microseconds); keep the shape/size/ready log lines; drop the value-range/mean lines; comment documents the by-construction guarantees and the pinning tests. Training-side twin at :1247-1254 (logging-only) untouched.

### LINKED ISSUE #401: Streaming inference: consume prefetched cadences in completion order
The streaming loop (main.py:855-869) consumes prefetch futures in strict catalog order (futures.popleft().result()) and refills one slot per consumed cadence. Measured consequence in inf_blpc3_20260807_011509: head-of-line blocking — when the oldest future sat in a 743 s load_lognorm, the other slots were finished-and-waiting, no new submission could enter, and preprocessing concurrency hit ZERO for 12% of the wall (1,511 s) while the 32-process ED pool and all 5 GPUs idled. Effective concurrency was 2.05 of a configured 3.

Change: consume whichever future completes first (wait FIRST_COMPLETED), submitting a replacement immediately on each completion. Per-cadence science is already order-independent (seeding keys on unit.index, the catalog position). To keep the run-level reproducibility contract (#282) intact under nondeterministic arrival order, replace the sequential algorithm-R reference-cloud reservoir with a random-key (bottom-k) reservoir whose keys derive from (root seed, STREAM_REFERENCE_CLOUD, cadence seed_key, row index) — the selected sample is then a pure function of the survey content, identical across arrival orders AND prefetch depths (strictly stronger than today, where the sample is depth-invariant only because order is forced). Manifest/DB row ORDER becomes nondeterministic (content unchanged); docs updated accordingly.

Note: the optimal --prefetch-depth is being grid-searched under the current strict-order loop; after this lands the grid needs re-running (tracked separately).

### LINKED ISSUE #402: Docs: record per-cadence stamp-cap as considered-and-rejected
Analysis of inf_blpc3_20260807_011509 showed per-cadence stamp counts spanning 168 (median) to 329,749 (max, BOL520 C-band at 8438 MHz), with the runtime heavy tail driven entirely by a handful of RFI-dense cadences. Capping stamps per cadence (top-K by detection statistic) would bound those tails.

Decision: REJECTED — a real signal co-located with dense RFI could rank below the cap and be dropped before the RF ever scores it; sensitivity in RFI-dense cadences is exactly what a galaxy-survey search cannot silently trade away. Secondary: a cap changes the ED fingerprint, invalidating every existing stamp cache. Record this in docs/INFERENCE_PIPELINE.md so the option isn't re-litigated from scratch each time the tail bites.

### LINKED ISSUE #403: ARCHITECTURE.md describes the inference prefetch as a 1-worker thread
docs/ARCHITECTURE.md:117 says the streaming CSV inference path uses "a 1-worker preprocessing prefetch thread". Stale: the implementation (main.py:839-869) is a ThreadPoolExecutor with max_workers = inference.prefetch_depth (default 3) and a depth-bounded in-flight window. docs/INFERENCE_PIPELINE.md describes it correctly. Update the wording to match the implementation as it stands after the completion-order change, and sweep docs/ for any other single-prefetch phrasing.

zachtheyek:
Closes #394
Closes #395
Closes #396
Closes #397
Closes #398
Closes #399
Closes #400
Closes #401
Closes #402
Closes #403

Ten issues from the deep-dive analysis of inference run `inf_blpc3_20260807_011509` (350-cadence subset, 3h29m wall, 569 candidates), grouped into one PR: the review-surface half attacks the RF false-positive triage problem, the throughput half attacks the measured pipeline stalls.

## Review surface

- **#394 — candidate frequency map**: new viz figure, one dot per candidate at (frequency, target row) colored by band. Vertical alignment across targets is the terrestrial-transmitter signature no other figure showed (49% of the run's candidates sat in two target/bands; several top candidates in GPS/Iridium allocations).
- **#395 — report-time frequency exclusion**: `--report-exclude-frequency-range START END` (repeatable) + config field. Strictly report-time: detection, `inference_results` rows, and saved figures untouched; Slack uploads, gallery membership, and tallies filtered, with original vs excluded vs reported counts in the run summary and summary card. Key joins both fingerprint denylists so it can never stale a resume row.
- **#396 — `utils/candidate_rfi_report.py`**: standalone, read-only, stdlib+numpy triage report (perband_report pattern): multi-target frequency coincidence, known-allocation flags (GPS/GLONASS/Iridium), per-target concentration, exclusion parity. Flags only, never deletes.
- **#397 — OOD triage**: reference cloud now persists its latent rows (+`n_contributing_cadences` representativeness signal); candidates get Mahalanobis OOD scores vs the survey cloud and vs the training run's true-class features (display-tag-aware artifact lookup, z_mean variant only, graceful skips); `candidate_triage_{tag}.csv` in review order; gallery/per-candidate ordering becomes confidence desc → survey-OOD desc → MC spread asc, with OOD percentiles on the gallery titles. Scores rank review order and never gate candidacy — a genuine technosignature is itself OOD w.r.t. synthetic training data.

## Throughput

- **#401 — completion-order consumption**: the streaming loop consumes whichever prefetch future finishes first instead of strict catalog order. The depth-3 baseline measured 12% of wall (1,511 s) at ZERO preprocessing concurrency from head-of-line blocking behind stragglers (worst: a 743 s stamp load) while the 32-process pool and 5 GPUs idled. Per-cadence science is order-free (seeds key on the catalog index); the reference-cloud reservoir is reworked from sequential algorithm R to a content-keyed bottom-k so the persisted cloud is byte-identical across arrival orders and depths. Manifest/DB row order and the latent-projection figure's bounded non-candidate subsample become order-varying (documented); all other figures render from catalog-sorted records.
- **#400 — load_lognorm sanity passes**: the full-array min/max/mean triple (3 of 13 passes over up-to-65 GB arrays; ~400-550 s per subset run, and the head-of-line stalls' biggest ingredient) is replaced by a row-0 spot-check with identical raise semantics — made genuinely unreachable-by-construction by extending the upstream validity filters to reject negative finite values (the one input class whose log NaN-poisoned a row past the old filters).
- **#399 — stamp-cache pruning default OFF**: the old AUTO default deleted every stamp the moment it was scored, so re-scores under the same ED config re-paid the measured bulk of wall time (25.8k stage-seconds). Stamps are now retained by default (~1 GB/cadence; free re-scores); catalog-scale runs must opt in with `--prune-stamps` (loud run-start reminder; trade-offs in flag help, config comment, docs, README).

## Hygiene

- **#398** — the "Memory usage: 0.00 GB" lines were the loaded array's nbytes under a misleading label and flooring format; renamed to `Cadence/Background array size` in MB.
- **#402** — per-cadence stamp caps recorded in docs as considered-and-REJECTED (sensitivity in RFI-dense cadences is not silently tradeable; caps also invalidate every ED cache).
- **#403** — ARCHITECTURE.md's stale "1-worker preprocessing prefetch thread" corrected to the depth-N completion-order pool.

## Verification

- Full local suite (`pytest -m "not gpu and not cluster"`, tf-stub env): **1,283 passed; the 73 failures are byte-identical to master** (pre-existing local-stub limitations that pass in CI). ~150 new unit tests ship with the changes, including regression pins for the reservoir's order-independence goldens, the stdlib-only cli import contract, the render-cap starvation fix, and the negative-value validity rule.
- A five-agent adversarial audit (4 hunters + verifier) ran against the full diff before this PR; all 19 confirmed findings (2 blockers, 4 should-fix, 13 nits) are fixed in the final commit.
- README CLI reference regenerated (`print_cli_help.py`, sync test green). ruff lint + format clean.

Note for reviewers: a `--prefetch-depth` grid search is running on blpc3 against the v1.1.0 container to pick a new default under the OLD strict-order loop; per #401's issue text the grid re-runs after this merges, and any default change lands as a separate evidence-based PR.


zachtheyek:
Thanks for the first pass — all five inline notes, both sub-inline extras, and the suggested test gap are applied in d73425f. Nothing to rebut this round.

**What changed, note by note:**

1. **`_log_norm_chunk_vectorized` docstring + silent rejections** — the docstring now states the full validity rule including the #400 negative clause and explicitly marks it load-bearing on *both* sides of the bit-identity contract (so nobody "simplifies" one side back). The streaming caller no longer discards the valid mask silently: a rejected-row count now logs a warning, making the sensitivity loss observable instead of inferable from the funnel figure.
2. **Representativeness warning on resume** — reworded to name a resumed *or* retried pass as the expected cause and to read as interpretation guidance ("percentiles are drawn from that partial background") rather than an anomaly report.
3. **`--max-candidate-plots` semantics** — help text now documents the per-side cap (up to 2× with exclusions configured), the #397 review ordering, and the fact that Slack uploads stay bounded by the reported side alone. README CLI reference regenerated (sync test green).
4. **`candidate_rfi_report.py` exit code** — "no candidates" now exits 0 (a clean run is an informative outcome); the genuine failure cases keep exit 1. Test re-pinned accordingly.
5. **#398 unit choice** — adaptive: MB below 0.1 GB (small arrays never print 0.0), GB above (the budgeted 65 GB cadence logs `64.83 GB`, not `64830.0 MB`). Both the cadence and background lines share one helper.

**The two sub-inline extras:** survey-OOD cloud latents are now memoized by `(path, mtime)` so the second scoring pass (triage CSV after gallery) reuses the load; and the considered-and-rejected blockquote is indented into ordered-list item 6 with a blank line so GFM can't terminate the list.

**The test gap:** two new tests in `test_seeding.py` pin exactly the trap you flagged as comment-only — one asserts the `(S, 2, 0) == (S, 2)` trailing-zero identity is real, the other asserts the keyless `(S, 3)` stream actually diverges from catalog cadence 0's `(S, 2, 0)`.

CI is green on the new head. Over to you for a second pass: @claude

zachtheyek:
Both second-pass nits applied verbatim in 9c8f65d: the `_lognorm_worker` / `_downsample_worker` docstrings now state the post-#400 rule (so the "load-bearing on BOTH sides" claim is backed on both sides), and the rejected-row warning is symmetric on the pooled/legacy collector path. Unit tests added for `_format_array_size` (no floored-to-zero zone at either end) and the streaming-load rejection warning. Merging once CI confirms green on this head — closing out per the post-LGTM convention, no re-tag needed.

PR #407, opened 2026-08-08
Raise inference.prefetch_depth default to 4 (grid-search evidence)

zachtheyek:
Closes #406

Changes `inference.prefetch_depth` default 3 → 4, with the evidence trail in the config comment, CLI help, docs/INFERENCE_PIPELINE.md step 4, and the regenerated README CLI reference.

## Evidence

Two grid searches on blpc3 (subset_1day_350.csv, 350 cadences, cold stamp cache each run, byte-identical science every run — 332 / 1,936,094 / 569):

| loop | depth 3 | depth 4 | depth 5 |
|---|---|---|---|
| pre-#401 strict-order | 12,561 s | 11,427 s (**−9.0%**) | — |
| post-#401 completion-order (current master) | 10,227 s | **9,755 s (−4.6%)** | 10,420 s (+6.8% vs 4) |

Depth 4 beat depth 3 in both loops independently; depth 5 regressed under completion-order consumption (which removed the head-of-line stalls extra depth existed to mask). Observed RAM peaks: 207 GB (depth 3) → 215 GB (depth 4); worst-case budget note updated to ~5 in-flight cadences (~325 GB of 503), tunable down via `--prefetch-depth` on smaller hosts.

Honesty note (recorded in the config comment): the post-#401 −4.6% alone sits inside the documented ~10% cluster run-to-run noise band — the call rests on the sign agreeing across both grids, not on any single measurement.

No science or resume impact: `prefetch_depth` is fingerprint-denylisted and per-cadence outputs are depth-invariant by construction (catalog-index seeding + content-keyed reservoir, #401).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #409, opened 2026-08-09
Catalog-derived RAM preflight warning at inference startup

### LINKED ISSUE #408: RAM-aware prefetch-depth preflight warning at inference startup
Follow-up from the #407 review discussion. The prefetch depth-4 default's RFI-dense worst case (~5 in-flight cadences x ~65 GB ≈ 325 GB) can exceed the README's advertised ≥288 GB RAM floor, and main.py currently builds the prefetch pool with no RAM-aware preflight — a 288-320 GB host hitting a run of RFI-dense C-band cadences OOMs rather than stepping down.

Design constraints agreed in the #407 review:

- **Warn, don't silently clamp.** A silent step-down would make wall-clock comparisons across hosts quietly incomparable (and would have muddied exactly the kind of grid search #406/#407 rest on). A startup log line — e.g. "depth 4 x ~65 GB worst case ≈ 325 GB vs 288 GB detected; consider --prefetch-depth 3" — gives the operator the protection without moving the ground under benchmarks. If a clamp is ever wanted, gate it behind an explicit opt-in flag.
- **The trigger is a worst case, not a measurement.** The ~65 GB figure is the RFI-dense C-band tail (per-cadence stamp counts span 168 median → 329,749 max), so a preflight keyed on the static worst case fires on nearly every sub-400 GB host, including catalogs containing no such cadence. Decide up front whether it keys on the static worst case (simple, noisy) or on something catalog-derived (e.g. a per-band estimate from the catalog's band mix).

zachtheyek:
Closes #408

Implements the RAM preflight per the design constraints recorded in the issue (from the #407 review discussion):

- **Warn, don't clamp**: the run proceeds untouched; the warning names the arithmetic and suggests a `--prefetch-depth` that fits the budget. A silent step-down would make cross-host wall-clock comparisons quietly incomparable.
- **Catalog-derived trigger**: the estimate is `(depth + 1) in-flight cadences × the largest per-band worst-case footprint among the bands actually present in the pending catalog` (C ~65 GB / L ~15 / S ~12 / X ~8 — observed tails from the 350-cadence subset benchmark, provenance in the constants comment). An X-band-only catalog stays quiet on a 288 GB host where the static C-band figure would cry wolf; unknown bands and catalogs grouped without a `Band` column assume the conservative C figure. Threshold: ~90% of host RAM (remainder covers TF/models/pool overhead and page-cache headroom, matching the README's OOM note).

The preflight is a pure function (`_prefetch_ram_preflight`) unit-tested across the warn/quiet/conservative/degenerate matrix, with psutil deferred and the whole call exception-guarded — it can never fail a science run.

Docs (decided per the issue scope): README RAM bullet gains the startup-warning clause; INFERENCE_PIPELINE.md step 4 documents the estimate and warn-only rationale; BENCHMARKING.md cross-references the preflight as the pre-run complement to its post-run RAM-pressure suggestion rule. No CLI change, so no README CLI-reference regeneration needed (sync test green).


zachtheyek:
All four notes and the three nits applied in the new head, nothing to rebut:

1. **Case-insensitive band lookup** — now matches `derive_cadence_provenance`'s treatment of user-supplied headers; a catalog headed `band` resolves normally instead of falling to the conservative branch. Lowercase test added.
2. **depth==1 edge** — when no step-down fits the budget, the message now says so honestly ("no --prefetch-depth fits this host's budget for band C — a smaller catalog slice or a larger-RAM host is the real fix") instead of suggesting the depth the user is already at. Pinned by a dedicated test.
3. **Import rationale + units** — the comment now states the real reason for the local import (psutil is a hard transitive dep via the monitor; the guard exists so the preflight can never fail a run), and the warning prints the host total in both decimal GB and GiB with a note that the GiB figure is what the monitor's "GB" line reports — the two log lines now reconcile instead of disagreeing by ~7%.
4. **Test coverage** — the fake `SimpleNamespace` units are replaced with real `PendingCadence`/`CadenceGroup` objects, and a new caplog test drives the actual call site through `_run_streaming_csv_inference` on a monkeypatched 100 GB host, pinning the argument wiring and the "can never fail a science run" contract against silent degradation.

Nits: the message names the driving band explicitly; `pending` is typed `list[PendingCadence]`; the `(#408)` prefix is gone from the Slack-bound text.

CI re-running on the new head. Second pass when ready: @claude

zachtheyek:
The half-fix is now a whole fix — your arithmetic was exactly right. The candidate is tested unclamped (`1 <= suggested_depth < depth`), so the entire nothing-fits family — not just `depth == 1` — lands on the honest "no --prefetch-depth fits" message; a regression test pins your precise 100 GB / depth-4 instance (budget 90, raw candidate 0, old code suggested a depth budgeting 130 GB). The rendering nit is fixed too: no-band-column catalogs now read "driven by the conservative unknown-band worst case (no band in the grouping)" instead of "band ? of ?", with a test. CI re-running on the new head. Over to you: @claude

PR #417, opened 2026-08-09
Content-addressed stamp cache + unified cache root under data_path/cache

### LINKED ISSUE #412: Content-addressed stamp cache + unified cache root under data_path/cache/{stamps,pfb}
Two coupled changes to the preprocessing caches, from the cache-architecture review discussion:

## 1. Content-addressed stamp-cache identity

Today's key bakes the catalog CSV's basename stem into both the cache directory (`{data_path}/inference/preprocessed/<csv_stem>_ed<hash12>/`) and every per-cadence filename (`{csv_stem}_{sanitized_group_key}.npy`). A cadence's stamps are a pure function of exactly two things — the ORDERED list of its h5 paths and the ED config — so the stem-based key causes spurious misses the resume guard never gets to save: the same files under a renamed/subset/superset catalog re-preprocess from scratch, edited group-key metadata (a fixed Target typo) orphans the cache entry, and two CSVs in one run can never share stamps. For the model-iteration loop (same files repeatedly re-sliced into different CSVs) these misses cost ~3 h a piece where a warm cache costs ~46 min.

Change the identity to content-addressed:
- Directory: ED-scoped, catalog-free — `{data_path}/cache/stamps/ed_<hash12>/` (see part 2 for the root).
- Filename: `<sha-of-the-ordered-h5-path-list>.npy` (hash the paths IN CSV ORDER — ABACAD order is scientifically meaningful; catalogs listing the same files in a different order are different cadences and must miss).
- The metadata `.json` sidecar keeps the human-readable provenance (group key, source CSV, target) and remains the resume guard's evidence; the guard's existing `h5_paths` + `ed_config_fingerprint` verification becomes verification of exactly what the key encodes.
- **New (size, mtime) guard**: record each h5's `(size, mtime)` in the sidecar at extraction; the resume guard compares on reuse and warns + re-extracts on mismatch — closing the shared blind spot where a re-staged/repaired h5 at the same path silently serves stale stamps. Keep size/mtime OUT of the hash (benign re-stages must not churn keys); the guard-side check provides the safety.
- The `unique-basename-stems` constraint in `plan_cadences` (and its ValueError) becomes obsolete — remove it.
- Housekeeping: `inference_cadences.npy_path` rows already map tag → cache entries for per-run cleanup.

## 2. Unified cache root: `data_path/cache/{stamps,pfb}`

Move both caches under one conceptual home:
- Stamp cache: `{data_path}/inference/preprocessed/...` → `{data_path}/cache/stamps/...` (stays on data_path: hundreds-of-GB, input-like, must survive output cleanup).
- PFB response cache: `{output_path}/cache/pfb/` → `{data_path}/cache/pfb/` (tiny, parameter-derived, self-verifying — the move is for coherence).
- Update every path constructor, default, log line, test, and doc that references the old locations (INFERENCE_PIPELINE.md, PREPROCESSING.md, README disk notes, config comments, CLI help for --preprocess-output-dir, BENCHMARKING/GPU_RUNTIME if they cite paths). `--preprocess-output-dir` keeps its pin-an-explicit-dir semantics.

## Rollout / acceptance (on blpc3)

- Remove the old caches (`rm -rf` the legacy preprocessed dirs and old pfb cache after the mv) — one-time cold start.
- Re-run inference on `subset_test.csv` and `subset_1day_350.csv`; verify: cold run populates `data_path/cache/stamps/ed_<hash>/` with content-hashed names + sidecars carrying (size, mtime); a second run under the same ED config skips all energy detection (warm-cache re-score, ~46 min on the 350 subset); a renamed copy of the catalog ALSO hits the cache (the whole point); science outputs identical across cold/warm.
- Verify the (size, mtime) guard: touch one cached cadence's h5 mtime → that cadence alone re-extracts with the warning.

zachtheyek:
Closes #412

Re-keys the stamp cache to its true content identity and unifies both preprocessing caches under one root, per the issue's design:

- **Content-addressed stamp cache**: `{data_path}/cache/stamps/ed_<fingerprint12>/<sha12-of-ordered-h5-paths>.npy`. A cadence's stamps are a pure function of exactly (ordered h5 path list, ED config), and the key now encodes exactly that — the catalog's basename stem appears nowhere (it was baked into both the directory and every filename), so renamed/subset/superset catalogs over the same files hit the same entries instead of re-preprocessing ~3 h a piece. Path-list **order stays in the hash** (ABACAD order is scientifically meaningful; same files reordered must miss), pinned by an order-sensitivity test plus a golden-digest test that makes the exact digest a deliberate cross-version contract rather than a refactor casualty.
- **(size, mtime) guard**: the sidecar records each h5's `(size, mtime)`, snapshotted **before the cadence's first h5 read** (recording after extraction would let a mid-extraction re-stage record the new file's stats against stamps holding pre-restage bytes, permanently certifying them). The resume guard compares on reuse and warns + re-extracts on mismatch — closing the stale-stamps-after-restage blind spot — while keeping size/mtime **out of the key** so benign re-stages never churn it. Guard semantics are deliberately fail-open: an *un-stat-able* h5 (archived raws, container without the raw-data bind) degrades to warn-and-reuse — cannot-verify is not verified-stale, and the documented no-raw-bind warm re-score workflow depends on it; only positive evidence (mismatch, malformed sidecar) re-extracts, and type-malformed `h5_file_stats` are contained to a warn + re-extract (the guard's call site sits outside per-cadence containment).
- **Obsolete constraint removed**: the unique-basename-stems `ValueError` in `plan_cadences` (and its test) is gone — nothing is stem-keyed anymore. In its place, `plan_cadences` **dedupes by content identity**: overlapping catalogs in one run listing the same cadence plan it once (identical cadences produce identical stamps and scores; two units sharing one npy_path would double work and let `--prune-stamps` race the duplicate's prefetch).
- **Unified cache root**: the PFB response cache moves `{output_path}/cache/pfb/` → `{data_path}/cache/pfb/` (tiny, parameter-derived, self-verifying — moved for coherence). `--preprocess-output-dir` keeps its pin-an-explicit-dir semantics.
- **Two consequential cleanups**: the pre-#298 fixed-name tmp sweep entry is removed (no historical writer ever produced `<stem>.tmp.npy` next to a content-hashed stem), and the `--bandpass-debug-plot` PNG is named after the plotted ON file instead of the now-opaque stamp basename (matching its suptitle; it's a human-facing debug artifact).

Docs swept: INFERENCE_PIPELINE (cache paragraphs, artifacts table, PFB section, plus a **migration note** — the old trees are orphaned and should be deleted by hand, and a `--save-tag` started pre-#412 must not be resumed across the upgrade since resume/supersede key on `npy_path`), PREPROCESSING, ARCHITECTURE (data-flow node + directory trees), config comments, `--preprocess-output-dir` / `--prune-stamps` help, and the regenerated README CLI inference block (sync test green).

**Testing**: `pytest -m "not gpu and not cluster"` under the local tf-stub harness: 1200 passed, 73 failed — the failure list is byte-identical to a clean master baseline run (pre-existing local tf-stub limitations; zero new failures). New/updated unit tests: content-hash naming (determinism, order sensitivity, catalog-name freedom, concatenation non-ambiguity, golden digest), the (size,mtime) guard matrix (match, size-only mismatch with restored mtime, mtime-only, unreachable-h5 reuse + not-masking-a-real-mismatch, malformed length/type containment, legacy sidecar degradation), sidecar stats recording in the end-to-end extraction test, an end-to-end re-extract-on-mismatch test, planned-unit dedupe, and the PFB cache's new location. Cluster verification (cold/warm/renamed-catalog/touch-guard on blpc3 per the issue's acceptance list) follows post-merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1060,14 +1060,16 @@ options:
                         overlap-search stamps (default: 0.5)
   --preprocess-output-dir PREPROCESS_OUTPUT_DIR
                         Directory for per-cadence .npy outputs from
-                        preprocessing. Default: a per-CSV directory {data_path
-                        }/inference/preprocessed/<csv_stem>_ed<hash>/ keyed on
-                        the energy-detection config fingerprint — runs sharing
-                        an ED config reuse each other's stamps automatically,
-                        and any ED-config change resolves to a fresh
-                        directory. Pass a directory explicitly to pin/share
-                        one location (reuse is still guarded by the sidecar's
-                        recorded h5 paths and ED fingerprint)
+                        preprocessing. Default: the content-addressed stamp
+                        cache {data_path}/cache/stamps/ed_<hash>/ keyed on the
+                        energy-detection config fingerprint, with filenames
+                        hashed from each cadence's ordered h5 path list — runs
+                        sharing an ED config reuse each other's stamps
+                        automatically regardless of catalog name, and any ED-
+                        config change resolves to a fresh directory. Pass a
+                        directory explicitly to pin/share one location (reuse
+                        is still guarded by the sidecar's recorded h5 paths,
+                        ED fingerprint, and per-file size/mtime)
   --prune-stamps, --no-prune-stamps
                         Delete each cadence's stamp .npy right after its
                         'inferred' manifest row lands, keeping the metadata
@@ -1077,7 +1079,7 @@ options:
                         stamps are retained (~1 GB/cadence average) so re-
                         scoring the same data under the same energy-detection
                         config (new weights, threshold sweeps) skips
-                        preprocessing entirely via the fingerprint-scoped
+                        preprocessing entirely via the content-addressed
                         cache. Pass --prune-stamps on catalog-scale runs:
                         without pruning a full catalog writes ~30-90 TB of
                         stamps and dies on disk.
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -94,7 +94,7 @@ flowchart TB
         CSV["CSV catalog(s)<br/>{data_path}/inference/"] --> GROUP["group_observations_from_csv()<br/>rows → 6-obs cadences"]
         H5[".h5 filterbank files"] --> ED
         GROUP --> ED["energy detection per cadence<br/>DC spike → bandpass flatten → k² threshold"]
-        ED --> NPY["per-cadence stamp .npy + metadata .json<br/>preprocessed/&lt;csv_stem&gt;_ed&lt;hash12&gt;/"]
+        ED --> NPY["per-cadence stamp .npy + metadata .json<br/>cache/stamps/ed_&lt;fingerprint12&gt;/"]
         NPY --> ENC["InferencePipeline.run_inference()<br/>encoder → latents → RF P(true)"]
         ART -. "trained models" .-> ENC
         ENC -. "inference_results (positives) + inference_cadences manifest" .-> DB
@@ -236,10 +236,12 @@ Three roots, set by `AETHERSCAN_{DATA,MODEL,OUTPUT}_PATH` (defaults under
 {data_path}/
 ├── training/                          # background plate .npy files (config.data.train_files)
 ├── testing/                           # preprocessed test .npy files (config.data.test_files)
-└── inference/                         # CSV catalogs (config.data.inference_files)
-    └── preprocessed/<csv_stem>_ed<hash12>/ # per-cadence stamp .npy + metadata .json
-                                       #   (ED-config-fingerprint-scoped: runs sharing an
-                                       #    energy-detection config reuse each other's stamps)
+├── inference/                         # CSV catalogs (config.data.inference_files)
+└── cache/                             # unified preprocessing-cache root (#412)
+    ├── stamps/ed_<fingerprint12>/     # content-addressed per-cadence stamp .npy + metadata
+    │                                  #   .json — <sha12-of-ordered-h5-paths>.npy: runs and
+    │                                  #   catalogs sharing an ED config reuse each other's stamps
+    └── pfb/pfb_response_*.npy         # content-addressed PFB passband responses (tag-independent)
 
 {model_path}/
 ├── vae_encoder_{tag}.keras            # final encoder (the inference model)
@@ -260,7 +262,6 @@ Three roots, set by `AETHERSCAN_{DATA,MODEL,OUTPUT}_PATH` (defaults under
 ├── inference_reference_cloud_{tag}.npz # MC-scored survey reference cloud (candidate uncertainty plot)
 ├── db/aetherscan.db                   # SQLite (WAL) — all stats/results tables
 ├── logs/aetherscan_{tag}.log          # this run's log (mode="w": overwrites the same tag's log on rerun)
-├── cache/pfb/pfb_response_*.npy       # content-addressed PFB passband responses (tag-independent)
 ├── round_data/{tag}/round_XX/         # per-round training memmaps (deleted after the round trains)
 │   └── rf/                            #   plus the RF training dataset
 └── plots/
--- a/docs/INFERENCE_PIPELINE.md
+++ b/docs/INFERENCE_PIPELINE.md
@@ -55,25 +55,28 @@ from `inference.cadence_h5_path_col` (default `".h5 path"`):
   raises `KeyError` for that CSV (logged, and the CSV is skipped).
 
 `DataPreprocessor.plan_cadences()` turns the valid groups into `PendingCadence` work units,
-each paired with a deterministic output path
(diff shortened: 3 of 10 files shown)


zachtheyek:
Applied the one first-pass note in c8d9c17: the plan-time dedupe now keys on the ordered h5 lists themselves (`planned_npy_paths` is a `dict[str, list[str]]`), skipping only true duplicates; a same-name/different-files truncated-digest collision raises a `ValueError` naming the cadence and the fix direction instead of silently dropping science data. Both branches are unit-tested — the collision branch by monkeypatching `_cadence_npy_filename` to a constant, since a real 48-bit sha256 collision isn't constructable in a test. The reviewer's framing was right: pre-#412 a collision was self-healing (h5_paths guard → mutual re-extraction, wasteful but never wrong), and the dedupe had turned it into the one silent-drop path in the planner.

No rebuttals — the note was correct as written.

Local suite after the fix: `pytest -m "not gpu and not cluster"` under the tf-stub harness, 157/157 in test_preprocessing.py and the full-suite failure list remains byte-identical to the master baseline (73 pre-existing tf-stub failures, zero new).

@claude

zachtheyek:
Follow-up: I initially only saw the first of the two first-pass notes (they landed as two separate review submissions 18s apart — my reply above covers only the dedupe/collision note). The second note is now also applied in e5522f6: the unreachable-h5 branch collects the un-stat-able paths and emits ONE aggregated warning per cadence (count + first path + npy_path) instead of one per file, keeping the fail-open reuse and mismatch semantics unchanged. The unit test now pins exactly one aggregated warning (with the n/6 count) for a cadence with two missing raws. Both first-pass notes are therefore addressed; no rebuttals.

@claude

zachtheyek:
All second-pass dispositions settled in c7ba932 (plus e5522f6, which the second pass raced):

**Applied:**
- **Note 2b (document the dedupe)**: the plan_cadences paragraph in INFERENCE_PIPELINE.md now describes the dedupe where the removed unique-stems ValueError used to be — explicitly covering the same-CSV shape (same six files at two hit frequencies; `Frequency` sits in the group-by cols but energy detection re-derives hits over the whole band and never reads it, so the collapsed rows' stamps and scores are identical by construction), the first-occurrence-wins bookkeeping (one manifest row, one key, one candidate set), and the collision `ValueError`.
- **Note 3 (migration)**: the migration note now carries the re-key-without-re-extraction recipe — every old sidecar records `h5_paths`, so the new name is `sha256(json.dumps(h5_paths).encode()).hexdigest()[:12] + ".npy"` — plus the explicit `--preprocess-output-dir` orphaning callout. I stopped short of a `utils/` script: the sanctioned rollout for this repo's own clusters (issue #412's acceptance section) deletes the old caches for a one-time cold start, so the script would ship untested against its only real use case; the recipe preserves the option for anyone with a cache worth saving.
- **Note 4 (warning volume)**: already applied in e5522f6 — one aggregated warning per cadence (count + first path), unit-tested to exactly one record with the n/6 count.
- **Note 5 (path-string contract)**: one sentence added — keys compare path strings byte-for-byte, deliberately no `normpath`/`realpath` (resolution would break the unreachable-raws re-score case), so symlink aliases/mount-prefix changes miss by design.
- **Test gaps**: same-CSV duplicate case added; first-occurrence-wins now asserted by group key in both dedupe tests; `unit.index` contiguity across a skip pinned.

**Left as-is, with reasons:**
- **Note 6 (unconditional makedirs)**: agreed it's cosmetic — an empty `ed_<hash>/` after a run whose CSVs all failed to parse is harmless, and eager creation keeps the cache-dir-exists invariant for every downstream writer without re-deriving it per CSV.
- **Collision exception type** (inline follow-on): `NonRetryableInferenceError` is defined in `main.py`, and `preprocessing.py` can't import it without a cycle (`main` imports `preprocessing`). The `ValueError` reproduces the removed unique-stems constraint's exact behavior, the condition is ~2e-5-probable at 100k cadences, and burning the retry budget on it costs minutes against a wrong-science alternative — leaving it, happy to revisit if the exception ever moves to a shared module.

Local suite: 158/158 in test_preprocessing.py; full tf-stub suite failure list remains byte-identical to the master baseline (73 pre-existing, zero new).

@claude

zachtheyek:
Applied the three post-LGTM cosmetic notes verbatim in 320fc93: the pruning-race limitation now states its #412-widened scope (every catalog shares the default dir, so concurrent overlapping catalogs are covered, not just reruns of one), the dedupe paragraph notes the collapse extends across runs under one --save-tag (resume keys on npy_path — the shared cadence's only manifest row names the first catalog's key), and the migration recipe points out that <fp12> is the same 12-hex fingerprint already sitting in the old directory's name. Review loop closed — merging once checks are green, with the cluster acceptance run (cold/warm/renamed-catalog/touch-guard on blpc3) to follow per the issue.

Issue #416, opened 2026-08-09
One-time supply-chain check: pip freeze diff of the cached blpc3 .sif vs the published GHCR image


zachtheyek:
requirements-container.txt uses ranged pins, so image rebuilds on different dates can pull different installed versions with no requirements change — and the local cached .sif on a cluster is only assumed, never verified, to match the published image.

One-time experiment on blpc3: pull the aetherscan image fresh from GHCR into ~ (as its own .sif, without disturbing the cached ~/Aetherscan/aetherscan-ngc25.02.sif), run pip freeze inside BOTH images, and diff. If identical: delete the freshly pulled artifact from ~ and close this issue with a "verified identical" comment — nothing else to do. If they differ: keep both artifacts, record the exact version deltas in this issue, and propose next steps (at minimum: identify whether the cached or published image is the one the v1.1.0 release blessing actually ran against, and whether any delta is behavior-relevant or supply-chain-suspicious).

(Adapted from the pre-v1.1.0 review-thread suggestion via the retired handoff doc.)

zachtheyek:
**Verified NOT identical** (2026-08-09, blpc3). Pulled `ghcr.io/zachtheyek/aetherscan:v1.1.0` fresh into `~/aetherscan-416-check-v1.1.0.sif` (tag chosen = the current release, matching what the cached image is assumed to be; GHCR tags at check time: `v1.0.0`, `v1.1.0`, `latest`, plus two `fp-*` fingerprint tags) and diffed `pip freeze` (sorted) between it and the cached `~/Aetherscan/aetherscan-ngc25.02.sif`. **12 deltas** (cached → published):

| Package | Cached (blpc3) | Published v1.1.0 |
| --- | --- | --- |
| GitPython | 3.1.54 | 3.1.58 |
| altair | 5.5.0 | 6.2.2 |
| annotated-doc | 0.0.4 | *(absent)* |
| astropy-iers-data | 0.2026.7.20 | 0.2026.8.3 |
| coverage | 7.15.2 | 7.15.3 |
| filelock | 3.32.0 | 3.32.2 |
| hf-xet | 1.5.2 | 1.6.0 |
| huggingface_hub | 1.21.0 | 1.22.0 |
| pillow | 11.3.0 | 12.3.0 |
| shellingham | 1.5.4 | *(absent)* |
| streamlit | 1.40.2 | 1.54.0 |
| typer | 0.25.1 | *(absent)* |

[...]

Issue #415, opened 2026-08-09
Rehearse the v9 migration RAM footprint before upgrading any large pre-v9 DB


zachtheyek:
The DB sets PRAGMA temp_store=MEMORY (db.py:728), so SQLite's index-build scratch space during schema migrations lives in RAM. The v9 migration builds two indexes over the biggest tables; on a catalog-scale DB (tens of GB) that scratch space could be many GB of RSS, and it has never been measured on a genuinely large database.

Status check (2026-08-09): blpc3's production DB is ALREADY at user_version 9 — the migration ran on 2026-08-05 while the DB was small (~MB), so no further v9 migration will occur on that DB and no action is needed there. This issue records the residual: before upgrading any LARGE pre-v9 database in the future (another node's DB, a restored backup, a fresh catalog-scale DB created on old code), rehearse the migration on a copy and watch process RSS, so the memory footprint is learned in a rehearsal rather than by OOM mid-index-build. If the rehearsal shows a scary number, the mitigation is file-backed temp storage for the migration step.

Close with findings whenever the first such upgrade is rehearsed, or close as not-applicable if no pre-v9 large DB ever materializes.

(Carried over from the pre-v1.1.0 review threads via the retired handoff doc.)

zachtheyek:
Session-verified 2026-08-09, closing per the status-check note in the issue body: blpc3's production DB was confirmed at `PRAGMA user_version = 9` before the #412 rollout runs (the v9 migration ran 2026-08-05 while the DB was ~MB-scale), and as expected **no schema migration occurred** during the first post-#412 inference runs — startup logs went straight from "Database schema initialized with WAL mode" into planning, with no migration stall and no anomalous RSS to record. The temp_store=MEMORY concern is therefore moot for this DB.

The residual stands as documented: before upgrading any genuinely LARGE pre-v9 database in the future (another node's DB, a restored backup, a catalog-scale DB created on old code), rehearse the migration on a copy and watch process RSS; if the number is scary, the mitigation is file-backed temp storage for the migration step. Re-open (or file fresh) if such a DB ever materializes.

PR #418, opened 2026-08-09
Weekly-automation stagger + SHAP-values-cache fingerprint (+ dot-plot & Slack-buffer fold-ins)

### LINKED ISSUE #413: Stagger the weekly automation crons and add concurrency groups
All three weekly automation workflows (claude-dependency-check.yml, claude-flaky-test-tracker.yml, claude-update-docs.yml) are scheduled at the identical cron (0 1 * * 1) and none carries a concurrency: block. Consequences: every Monday three Claude-invoking workflows wake simultaneously and compete for runners/API capacity, and two of them can file issues/PRs at the same moment and race each other; separately, a scheduler-delayed run and a manual workflow_dispatch of the same workflow can overlap with nothing to serialize them.

Change: stagger the crons (e.g. 01:00 / 02:00 / 03:00 UTC Mondays) and add a uniform concurrency: group per workflow (group keyed on the workflow name; a second copy of the same workflow queues or cancels rather than racing). Update docs/GITHUB_AUTOMATION.md's schedule table to the new times and document the concurrency convention. Note: docs/GITHUB_AUTOMATION.md and the repo-context skill both state Mondays 01:00 UTC — sweep both for the new schedule.

(Carried over from the pre-v1.1.0 review threads via the retired handoff doc.)

### LINKED ISSUE #414: SHAP-values-side content fingerprint (close the coherently-stale cache pair gap)
#359→#381 gave the SHAP CLUSTERING cache a content fingerprint: train.py persists _shap_content_fingerprint(shap_values) alongside the cached embedding/cluster_labels and refuses a cache whose recorded fingerprint doesn't match the SHAP matrix in hand (train.py:6901-6922). The SHAP-VALUES cache itself has no equivalent self-validation tying it to the model/data that produced it.

The uncovered failure mode is the coherently-stale pair: a leftover values-cache from an older model is reused, its clustering cache matches it (they are consistent with each other), and nothing notices that neither belongs to the current model. Mirror the clustering-side fix: fingerprint the values cache against its own inputs (model + data identity) so a stale pair can't slip through as a matched set. Update the training-pipeline docs' SHAP-cache description accordingly.

(Carried over from the pre-v1.1.0 review threads via the retired handoff doc.)

zachtheyek:
Closes #413
Closes #414
Refs #415 — session-verified 2026-08-09 and closed: blpc3's production DB confirmed already at `user_version 9`, and no migration (hence no `temp_store=MEMORY` RSS event) occurred during the first post-#412 inference runs; the rehearsal residual applies only to future large pre-v9 DBs.
Refs #416 — session-verified 2026-08-09, left open with findings: the cached blpc3 `.sif` and the published v1.1.0 GHCR image are **not** identical (12 deltas — the cached image is a Jul-22 local build predating the W31 security bumps, so it is the stale side; scientific stack byte-identical; swap proposal recorded in the issue).

The unified weekly-automation + SHAP-cache maintenance bundle, plus two maintainer-requested fold-ins:

**#413 — staggered crons + concurrency groups.** dependency-check stays Mon 01:00 UTC, flaky-test-tracker moves to 02:00, update-docs to 03:00 — the three Claude weeklies no longer wake simultaneously and race for runners/API capacity. The issue's CAUTION on update-docs' window computation was verified unfounded for the hour change: its anchor query filters on event type + conclusion and reads the previous successful scheduled run's `createdAt` — never the cron string — so the first staggered run covers a once-off ~2 h-longer window with **no gap**, scheduler delays keep being absorbed by construction, and the other two workflows carry no time-anchored logic beyond the ISO-week dedup key (safe at any Monday-UTC hour). Every weekly gains `concurrency: group: ${{ github.workflow }}` with `cancel-in-progress: false` (house style per `release.yml`): a delayed weekly and a manual dispatch of the same workflow now queue instead of racing, distinct workflows never share a group, and not cancelling matters specifically for update-docs — a cancel between issue-creation and the CLI-help embed step would strand the week's issue without its regenerated CLI blocks, unrecoverable within the ISO week. `docs/GITHUB_AUTOMATION.md` gets the new hours plus a scheduling/concurrency-conventions paragraph; the repo-context `SKILL.md` bullet is updated (note: `.claude/` is maintainer-writable only in CI-authored PRs — this PR is maintainer-authored, so the edit rides along).

**#414 — SHAP-values-cache input fingerprint.** Mirrors the #359/#381 clustering-side design on the values cache itself: `rf_shap_values_{tag}.joblib` now records an `input_fingerprint` — one sha256 over the eval matrix + labels and the persisted RF artifact's **file bytes** — verified after the existing shape guard, warn + recompute on mismatch. Hashing the RF file (re-dumped from the in-process model on every compute) rather than an in-memory object means plots-only reruns with no model loaded validate against the same canonical bytes the pool workers consume. Pre-#414 caches migrate via one recompute; an unverifiable fingerprint (RF artifact missing) degrades to warn-and-reuse — cannot-verify is not verified-stale, and a forced recompute would itself need the missing RF. This closes the coherently-stale-pair gap `_shap_cache_consistent`'s docstring used to disclaim: a values+clustering pair that is internally consistent but belongs to an older model can no longer be served as a matched set. Tests mirror the clustering-cache suite (hit / legacy-migration / different-model-same-shapes / different-val-contents-same-shapes / missing-RF-unverified-reuse); `TRAINING_PIPELINE.md` + `MODELS.md` now state both caches self-validate.

**Fold-ins (maintainer requests, this session):** `logger.slack_buffer_size` default 100 → 20 (smaller Slack batches flush more promptly; `RUNTIME_SERVICES.md` updated — no CLI flag exposes it, so no README regeneration). And the candidate frequency map becomes an actual dot plot: candidates sit at their frequency along x with seeded jitter on y (the axis encodes nothing), **target moves off the y-axis into color + legend labels** (top-20 by candidate count + one aggregate entry; 20 = tab20's distinguishable-palette size), and band is deliberately absent from the legend — x-position implies it. The RFI read inverts from vertical alignment to one frequency carrying many colors; filename, suite registration, exclusion shading, and skip paths unchanged.

**Testing**: `pytest -m "not gpu and not cluster"` under the local tf-stub harness: 1206 passed, 73 failed — the failure list is byte-identical to the master baseline (pre-existing tf-stub limitations; zero new failures). #414 is training-side: the new fingerprint tests run fully under the stub (they stub the SHAP pool/explainer); the real-Keras surfaces ride the normal CI matrix on this PR.


zachtheyek:
All five first-pass notes applied in 4c64cf8 — no rebuttals:

1. **Model-component invariant**: recorded as a NOTE on `_rf_artifact_digest` (byte-identical re-dump of the same forest holds today; a resume under a different `rf.n_jobs` — `RandomForestModel.load` mutates the estimator — or a newer sklearn/joblib re-pickles differently and costs one spurious, fails-safe recompute), and the mismatch warning now names re-pickling as a cause alongside stale-cache-under-a-reused-tag, so an operator isn't pointed at a model change that never happened. I stopped short of the structural tree digest for now — it trades the no-load streaming property for re-pickle stability, and the NOTE marks it as the revisit path if spurious recomputes ever materialize.
2. **Gray collision**: tab20's gray pair (slots 14/15) is out of the target rotation (`_CANDIDATE_FREQ_PALETTE`, 18 colors); the cap now derives from the palette length and a monkeypatched/lowered cap can never cycle into duplicate colors (`min` + strict-False zip over the palette). Docs updated to top-18.
3. **Stale 01:00**: the 8-day-fallback comment now says 03:00 (#413 stagger).
4. **`train_features` folded into the fingerprint** — it feeds the log-loss background subset, and the helper's identity claim is now literally true. Write site, read site, and the test helper all carry the new component.
5. **Queue semantics**: GITHUB_AUTOMATION.md now notes GitHub keeps at most one pending run per group (a third entry replaces the queued one), benign under the ISO-week dedup guards.

Also took the optional test suggestion: the overflow test now captures the legend off the figure en route to disk and pins the redesign's contract — cap + 1 target-labeled entries, legend title `target`, and no bare band-letter labels.

Local suite: test_train_utils + test_inference_viz — 133 passed, 5 failed, the failures being the pre-existing tf-stub `TestEncoderTrainedHeuristics` set (byte-identical to the master baseline).

@claude

zachtheyek:
Second-pass note applied in 3958545 — no rebuttals. The four-cell table was right: the degrade now keys on "can we recompute?" (`model_loaded`, shared with the compute-path guard). Row 3 (RF unreadable + model in memory) recomputes from the in-process model — which re-dumps the artifact — instead of serving a possibly-older-model cache; row 2 (pre-#414 cache + no model) reuses the legacy cache unverified instead of dying on the pointed RuntimeError, restoring the pre-#414 plots-only behavior. `test_missing_rf_artifact_serves_cache_unverified` split into the with-model (recomputes, artifact re-dumped, fresh fingerprint) and without-model (serves unverified) cells, plus a legacy+no-model test; comments, the class docstring, and TRAINING_PIPELINE.md stop claiming a recompute would need the missing RF. The buffer-protocol nit is applied too (`digest.update(payload)` on the ascontiguousarray, no `.tobytes()` copies).

Local: test_train_utils 91 passed / 5 failed — the pre-existing tf-stub TestEncoderTrainedHeuristics set, byte-identical to the master baseline.

@claude

zachtheyek:
Applied the three post-LGTM cosmetics verbatim in e78e0c5: both remaining pre-fold-in enumerations (the `_shap_cache_consistent` docstring and the MODELS.md artifact row) now say train + eval matrices + labels + persisted RF, and `_rf_artifact_digest`'s contract line describes the read side's two-way recompute-vs-degrade decision. Took the offered skip on the try/except/else style item — keeping the flat sentinel form. Review loop closed; merging once checks are green.

